PR #16, opened 2026-08-05
Toast teal fixes

h3xxit:
## Summary by CodeRabbit

* **New Features**
  * Library toasts now support neutral, success, and danger tones with tone-specific styling.
  * Demo banners now visually distinguish expired vs. active states.
  * Copy-related actions show explicit success or failure toast feedback.
* **Bug Fixes**
  * Error/conflict/decline/withdrawal and access request notifications now consistently use danger tone.
* **Style**
  * Updated notification banners to use shared design-system colors and Button primitives (including danger/tiny actions with loading spinner where applicable).
  * Refreshed protected-branch and pull-needed banner styling to consistent informational themes.


edujuan:
/bevel-review

Context for this run — the head moved twice since CodeRabbit's pass:

1. `424d9bd` split `ToastContext` / `useLibraryToast` out of `state/toast.tsx` into `state/toast.context.ts`, per CodeRabbit's `react-refresh/only-export-components` finding. Same provider/context convention the repo already uses (`git.context.ts`, `layout.context.ts`, `open-change-requests.context.ts`).

2. `b9dda14` merges `dev`. This needs a look — **`dev` rebuilt the same component while this branch was open**. PR #12 replaced the teal card with the prototype's ink pill and retired its four raw hexes, so `state/toast.tsx` conflicted whole. I resolved toward dev's pill (base branch's decided direction, and it came from the prototype), keeping this branch's tone API and dev's `toastDuration(msg)`, and dropped `Surface` for the toast since the pill has no border or shadow.

   The consequence worth reviewing: **tone had to move from the text to the plate.** On the inverted pill `--color-ok` over `--color-ink` is 2.4:1 and `--color-danger` is 1.9:1; canvas text over the three plates is 12.3:1 / 5.1:1 / 6.5:1. That inverts `Banner`'s `bg-*-soft text-*` convention, so it's a deliberate divergence rather than an oversight — but it's a design call made during conflict resolution, not one the PR author signed off on.

Build, typecheck, 2076 tests and `ds:check` all pass; `mergeable_state` is clean.

---

edujuan:
Addressed in `82bf887` — 2 of the 4 findings taken, 2 declined with reasons.

**Taken**

- **Tone lost during fade-out** (`toast.tsx`) — real, and worse than it reads: `transition-[opacity,transform]` doesn't cover `background-color`, so a danger toast *jumped* red→ink at the start of its exit rather than recolouring gradually. The toast now carries a `visible` flag instead of dropping to `null`, keeping the plate alive for the 200ms exit.

  One correction to the suggested patch: the `useRef` version doesn't lint here. `eslint-plugin-react-hooks` v7 has `react-hooks/refs`, and `TONE[toast?.tone ?? lastTone.current]` fails with *"Cannot access refs during render."* Holding it in state avoids that. The message is still cleared on expiry — the pill is transparent, not unmounted, so a lingering message stays readable to a screen reader after it's left the screen.

- **`WelcomePage` copy failure missing `danger`** — correct, and it was the only `COPY_FAILED_TOAST` call site without it; the two Library sites both pass `danger`. It arrived with `dev` after the tone existed, so it never got wired up. Fixed.

**Not taken** — both in `dev`'s onboarding code, untouched by this branch except an import path:

- **Double announcement on dismiss** (`ConnectAgentPill`) — deliberate, and the code says so at lines 62-66: the toast only exists inside the Library's provider, while the local live region survives the pill's unmount and is what keeps the confirmation from going silent on Knowledge. The suggested fix needs provider-detection to switch on, and none exists — `ToastContext`'s default is a silent no-op, indistinguishable from a real provider at runtime.

- **Dismiss button hidden on touch** — the `pointer-events-none` is the guard, not the bug. Lines 117-121: an invisible-but-tappable 18px target performing an *irreversible* server write is a trap on any device without hover, so not reaching it is the intended safe failure. The proposed `pointer-events-auto opacity-100` would leave the × permanently visible on every pointer type, which is the outcome that comment exists to prevent. Worth raising with the onboarding author as its own change if the touch gap matters.

Build, typecheck, 2076 tests and `ds:check` pass; lint unchanged at 154 pre-existing errors. `mergeable_state` clean against `dev`.

---

edujuan:
/bevel-review

Re-run: the go-ahead above was published against `b9dda14`, and the head has since moved to `82bf887`, which has no `bevel/architecture-review` status. Since that status is per-head-SHA and is meant to serve as the branch-protection check, the current head is unreviewed.

What changed since the reviewed commit — both from CodeRabbit's pass on the merge:

- `state/toast.tsx` — the toast now carries a `visible` flag instead of dropping to `null` on expiry, so its tone survives the 200ms fade-out. `transition-[opacity,transform]` doesn't cover `background-color`, so the previous reset made a danger toast jump red→ink at the start of its exit. The message is still cleared on expiry, since the pill is transparent rather than unmounted and a stale message would stay in the accessibility tree.
- `WelcomePage.tsx` — the copy-failure path now passes `danger`; it was the only `COPY_FAILED_TOAST` call site without it.

Two of CodeRabbit's findings were declined, both in `dev`'s onboarding code and untouched by this branch: the pill's double announcement and the dismiss button's `pointer-events-none` are each deliberate and documented in place — reasoning in the comment above.

---

PR #21, opened 2026-08-05
Fix/skills sidebar context menu

h3xxit:
## Summary by CodeRabbit

* **New Features**
  * Added right-click menus across the Library sidebar for groups, lenses, and empty navigation space.
  * Users can create groups, add items, copy links, and manage access directly from the sidebar.
  * Menus appear at the pointer location and close with outside click or Escape, restoring focus afterward.

* **Bug Fixes**
  * Improved handling for locked rows and selection state.
  * Copy-link actions now show clearer success and failure feedback.

### DIFF
--- a/.changeset/olive-donkeys-listen.md
+++ b/.changeset/olive-donkeys-listen.md
@@ -1,0 +1,30 @@
+---
+'@bevel-software/platform-core-frontend': patch
+---
+
+Right-click works in the Library nav, the way it already did in Knowledge's file
+tree. It answered nothing before — no `onContextMenu` existed anywhere in
+`GroupsSidebar`, so the gesture fell through to the browser's own menu on the
+one surface that most looks like it should have its own.
+
+The BEHAVIOUR is the file tree's, deliberately: the panel opens at the pointer,
+an outside click or Escape closes it, and Escape hands focus back to the row it
+came from — the same `useDismissableMenu` wiring, since `MenuPanel` is
+presentation only and provides none of it. Two sidebars in one frame should not
+answer the same gesture two different ways.
+
+The ITEMS are the Library's own, because a group is not a file. A group row gets
+`Add a skill or tool` · `New group` · `Copy link` · `Manage access`; a LENS row
+("Owned by me", your own space) gets only the two that a slice of the catalog
+can answer, since there is no folder and therefore no `access.md` behind a
+slice — the same call `PageActions` already makes when it hides `Share` on the
+personal page. The nav's empty space gets `New group` alone. Rename, Delete and
+Download are deliberately absent: no endpoint stands behind any of them for a
+group, and a menu item that cannot do its job is worse than one that is not
+there.
+
+`Manage access` stays ungated on `canWrite`, exactly as the group page's `Share`
+is — for a non-writer the dialog renders read-only, which is precisely what "who
+is this shared with?" should answer, and for an admin locked out of a group it
+is the self-service way back in. Locked rows get the full menu for the same
+reason they get the same click: a group you are not in is still a place.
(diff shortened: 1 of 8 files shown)


edujuan:
Pushed two commits, plus a merge of `dev` into this branch. The merge is the point, so it's worth saying why.

**The `test` check wasn't failing on anything in this branch.** `apps/web` was failing to build with `MISSING_EXPORT` on `useLibraryToast`, but the branch builds clean on its own — and so does `dev`. CI builds the *merge* commit (`actions/checkout` on `pull_request` resolves `refs/pull/21/merge`), and the break only exists there. A semantic merge conflict: git had nothing to flag, which is why the PR read as mergeable while the build was red.

`dev` (#16) split the toast channel out of `toast.tsx` into `toast.context.ts`, so the provider module exports nothing but a component and Fast Refresh keeps working. This branch predates that split, and `LibraryLayout`'s new import still pointed at `../state/toast`. Different files on each side, so the merge was textually clean and semantically broken.

Two things followed from it:

- The import now points at `../state/toast.context`.
- That split also widened the hook — `ShowToast` takes a tone now. A copy that *failed* was rendering on the same ink plate as one that succeeded, so the colour said "copied" while the words said otherwise. Toned the pair the way `AddToGroupDialog` and `PersonalAddDialog` already do: neutral on success, `danger` on refusal. The existing refusal test asserted the message but not the plate, so it passed either way; it now asserts the plate and fails against the pre-fix source.

CodeRabbit's `LINK_COPY_FAILED_TOAST` finding was valid and is applied as suggested — `onCopyLink` is gated on `menuFilter`, not `menuGroup`, so lens rows do offer `Copy link`, and the sidebar test asserts exactly that for "Owned by me" and the personal space. Worth noting the constant's own doc comment already said "open the row"; only the string had drifted.

Green locally on everything CI runs — `pnpm build`, `pnpm test` (913 frontend, 1222 backend), `pnpm ds:check` (no regressions) — plus `eslint` on the changed files and the `core-frontend` typecheck.

One call worth flagging: merging `dev` in was the way to make the failure reproducible and fixable, but it does put those commits on your branch. Rebase it away if you'd rather.

---

edujuan:
/bevel-review

Code-level pass is clean: CodeRabbit approved `199ea32` and `test` is green on that head.

For the record, the five nitpicks from the first review round — all verified against the code, none applied, each for a stated reason:

**Focus into the panel on open** and **clamp the panel to the viewport** (`GroupsSidebarMenu`) — both are real, and both are true of Knowledge's file-tree menu too. `ContextMenu` in `FileExplorer.tsx` positions with a bare `style={{ left: x, top: y }}` off raw `clientX/clientY` (L228/L535) and focuses nothing on mount. This component mirrors that menu on purpose — its docstring and the changeset both say so, and "two sidebars in one frame should not answer the same gesture two different ways" is the point of the PR. Fixing one menu and not the other trades a small edge-case bug for the inconsistency the change set out to remove. Worth doing as a shared positioning/focus helper across both menus; that's a separate change, not this one.

**`patch` → `minor` on the changeset** — checked the convention rather than the principle. Every pending changeset for `@bevel-software/platform-core-frontend` is `patch`, including `notification-design-system` and `menu-and-dock-layering`, which are feature-scale. CodeRabbit's own escape hatch ("if this package treats every sidebar change as a patch, keep it as is") applies.

**Reset the access mock per test** — the concern is right in principle: `test-setup.ts` clears localStorage but not mock call history, and the file deliberately avoids `restoreAllMocks` for a documented reason. Today `mock.calls[0]` is still correct, because only one test actually clicks `Manage access` — the earlier ones only assert the item is present. So it passes for a real reason now, but it would silently start passing for the wrong one the moment a second test clicks that item. Left alone as it's the author's file and CodeRabbit rated it trivial, but a `mockClear()` in the existing `beforeEach` would close it cheaply.

**Simplify the ref construction** — cosmetic, no behavioural difference.

**Docstring coverage (50%, ⚠️)** — not chased, and the number is not trustworthy. That run reported "Files skipped from review as they are similar to previous changes", so the percentage was reused rather than recomputed. Auditing the diff by hand: 4 of 6 top-level functions carry docblocks. Of the two that don't, `GroupsSidebarMenu` has a 26-line docstring plus a per-prop documented interface — it just sits on `GroupsSidebarMenuProps` rather than on the function, so the counter misses it — and `GroupsLabel` is a 12-line presentational label. `.coderabbit.yaml` doesn't configure pre-merge checks, so the 80% threshold is an org default. Adding a filler docstring beneath a better one to move a miscounted metric would make the file worse.

---

PR #29, opened 2026-08-06
Empty skill md instead of workspace switch

h3xxit:
## Summary by CodeRabbit

* **New Features**
  * Create groups directly with name validation and clear error messages.
  * Create personal or group skills in place with an empty `SKILL.md` ready to edit.
  * Skill creation supports direct saving or review submission based on permissions.
  * Personal spaces are created privately when needed and excluded from group listings.
* **Bug Fixes**
  * Prevented duplicate skill names, including case-insensitive and cross-group conflicts.
  * Added safeguards against accidentally replacing existing skill files.
* **Tests**
  * Expanded coverage for skill creation, validation, navigation, permissions, and group provisioning.


h3xxit:
Review triage for the last CodeRabbit round (72a7a3e):

**Fixed**
- Creations are now serialized by lowercased folder name (check + exclusive write under one per-name lock) — closes the `GTM`/`gtm` double-create on case-sensitive filesystems.
- `ensurePersonalGroup` treats a provision 409 as "already exists" and returns `created: false` — ensure semantics hold under cross-process races.
- `POST /groups` tolerates a missing JSON body (400 instead of a destructure throw).
- `NewGroupDialog` navigates with the server-returned folder name; the test mock now returns a differing folder so a regression fails.

**Skipped, with reasons**
- `existsAtRef` root-ref validation + atomic new-folder authorization: both target code this PR deletes (`existsAtRef` and the gate carve-out are gone; provisioning is the atomic door).
- `NewSkillPanel` error copy: the message already covers `.` and `..`.
- Skill-creation atomicity in `library.api`: the direct path already uses the exclusive `ifAbsent` create; the propose path writes to the author's own suggestion branch, where the change request review is the arbiter.
- Personal-prefix migration / legacy ungrouped bucketing: there are no production deployments — backwards-compatibility code is explicitly out of scope for this repo right now.
- Global cross-group skill-name uniqueness: pre-existing catalog-level concern (`dedupeById`), tracked separately from this PR.

PR #38, opened 2026-08-07
Deployment settings page: the setup screen, reachable after setup

h3xxit:
Gives the first-run setup form a permanent address at `/deployment` (settings nav, admin section), so single sign-on, the knowledge-base connection, and the branch model can be configured after setup — not only while the gate is open. The backend accepted these writes from any admin all along; only the door was missing.

- `SetupScreen` gains a `variant`: `'setup'` keeps first-run chrome, `'settings'` embeds the identical form (fields, env-lock rule, connection test, restart notices) in a host page.
- New `DeploymentPage` with admin gating (presentation only — the backend enforces the real gate), loading and retry states.
- The settings-nav allow-list derives from the row list, so the new row self-registers.

Validation: 3 new page tests, frontend 971/971, typecheck clean, web build green.

## Summary by CodeRabbit

* **New Features**
  * Added a Deployment settings page at `/deployment` for administrators.
  * Deployment settings support knowledge, skills, and tools connections, branch configuration, and SSO settings.
  * Added Deployment to settings navigation.
  * Empty repositories now default branch settings to `main`.

* **Bug Fixes**
  * Added loading, retry, and save status handling.
  * Non-administrators receive an access message without triggering settings requests.
  * Save errors preserve entered settings and display restart guidance.
  * Clarified setup guidance and automatic version-field population.

### DIFF
--- a/.changeset/setup-gets-an-address.md
+++ b/.changeset/setup-gets-an-address.md
@@ -1,0 +1,11 @@
+---
+'@bevel-software/platform-core-frontend': patch
+---
+
+Deployment settings get a permanent address. The first-run setup form —
+knowledge-base connection, branch model, single sign-on — is now also an
+admin settings page at `/deployment`, so adding SSO (or rotating its
+client secret, or renaming a branch) after setup no longer means finding
+the right environment variable. Same form, same env-lock rule, same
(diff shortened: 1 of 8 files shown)


h3xxit:
@coderabbitai full-review

PR #59, opened 2026-08-11
Give the review diff the document measure; stop skills switching apps

h3xxit:
Two reports against the re-mounted agent-changes panel.

## 1. The diff took the whole width

The same document reads at an 880px line in the Knowledge surface two clicks away, so the place someone reads most carefully had the worst measure in the app.

It now uses `DOCUMENT_COLUMN` + `documentGutters` — the **tokens**, not `KbDocumentShell`. The shell also owns the scroller, the rail track and the file lock's scroll listener, none of which belong to a floating panel with its own header and footer; the measure lives in `shared/theme` precisely so it can be used without the rest.

`MarkdownDiffViewer` gains a `scroll` prop, mirroring `KbMarkdownView`'s. Its root is `h-full overflow-auto … px-4 py-3`, so wrapping it in a column would have left the **outer** box unable to scroll and stacked both paddings — the nested-scroller trap this component's own doc comment warns about in the other direction.

The marked-source view for yaml, scripts and config stays full-bleed: there a line is a line of code.

## 2. Selecting a changed skill switched the whole app

`handleSelect` opens the file beside the diff for context. For a `Groups/…` path that URL is a **library location**, so the shell swapped to Skills & Tools and the panel, the diff and the review disappeared — because someone clicked a row in a list.

Group items are now selected without navigating. The diff is already on screen beside the row, which is the context that call exists to provide, and for a group item it is the only context that can be shown without leaving.

## One of these is untested, and I'd rather say so

I wrote three tests for the navigation change, then found the first **passed with the guard deleted** — the file picker doesn't open under the test harness, so every assertion held whether the fix existed or not. A second attempt asserting on the `openFile` call rather than the resulting location hit the same wall.

Rather than ship green tests that measure nothing, they're gone.

- **The `scroll` prop IS tested** — two cases in `MarkdownDiffViewer`'s own suite.
- **The group-item skip is reasoned but unguarded.** Worth a look on staging: select a changed skill in the review panel and confirm you stay in the panel.

## Gate

typecheck, build, **2420 tests** (1358 core-backend / 1062 core-frontend) and `ds:check` green. Released as **0.7.1** (patch).

---
## Summary by cubic
Aligns the review Markdown diff to the document column and stops the app from switching when selecting `Groups/...` changes. Bumps core packages to `0.7.1`.

- **Bug Fixes**
  - Wrap Markdown diffs with `DOCUMENT_COLUMN` to match the document measure; non‑Markdown diffs stay full‑bleed.
  - Keep the review panel open when selecting `Groups/...` changes by skipping `openFile` for group items and deleted files. Adds `isGroupItemPath` to normalize workspace‑relative paths, with unit tests.

- **Dependencies**
  - Bump `@bevel-software/platform-core-backend`, `@bevel-software/platform-core-frontend`, and `@bevel-software/platform-shared` to `0.7.1`.

<sup>Written for commit b03474cff000680a8148286942463f1eb99650c0. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Bevel-Software/Hexis/pull/59?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/core-backend/package.json
+++ b/packages/core-backend/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@bevel-software/platform-core-backend",
-  "version": "0.7.0",
+  "version": "0.7.1",
   "description": "Open-source core backend of the Bevel platform: git-backed knowledge workspace, workflow (branches/change requests/locks/SSE), skills, tools, secrets vault, access control and the remote MCP surface.",
   "type": "module",
   "main": "./dist/index.js",
--- a/packages/core-frontend/package.json
+++ b/packages/core-frontend/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@bevel-software/platform-core-frontend",
-  "version": "0.7.0",
+  "version": "0.7.1",
   "description": "Open-source core UI of the Bevel platform (raw TS/TSX source): workspace explorer/viewer, branches & change requests, Library, secrets, tools, access control.",
   "type": "module",
   "main": "./src/index.ts",
--- a/packages/core-frontend/src/modules/review/components/ReviewPanel.tsx
+++ b/packages/core-frontend/src/modules/review/components/ReviewPanel.tsx
@@ -7,8 +7,12 @@ import {
   useFileNav,
   useNodeIdNav,
   resolveRelativePath,
+  stripJunkBeforeKbDir,
   KB_ROUTE_PREFIX,
 } from '../../workspace/routing/kb-routes';
+import { groupOfPath } from '@bevel-software/platform-shared';
+import { cn } from '../../../lib/utils';
+import { DOCUMENT_COLUMN } from '../../../shared/theme/measure';
 import { ReviewFileRow } from './ReviewFileRow';
 import { DiffViewer } from './DiffViewer';
 import { MarkdownDiffViewer } from './MarkdownDiffViewer';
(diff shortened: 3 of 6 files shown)


h3xxit:
Reworked after review feedback — you were right that `MarkdownDiffViewer` is correct and the container is what's wrong.

**Before:** I added a `scroll` prop to `MarkdownDiffViewer` so the parent could own the column. That changed a component **three surfaces** share (this panel, the change-request dialog, file history) to fix a problem in one of them.

**Now:** the viewer is byte-identical to before. The wrapper constrains **width and nothing else** — `DOCUMENT_COLUMN`, no more — and the viewer keeps its own scroller and padding.

Two things that fall out of doing it this way, both deliberate:

- **Not `documentGutters`.** It bundles `pb-[110px]`, and the scroller is the *child* here — that bottom rhythm would sit outside it as permanent dead space, with the scroll ending short of the panel.
- **The line lands at 848px** (the child's own `px-4` becomes the gutter) against Knowledge's 800px. Near enough that the two read as the same column, and bought without reaching into shared code.

Diff is now **one component file** instead of three:

```
ReviewPanel.tsx        +52 −12
ReviewPanel.test.tsx    +5 −1   (kbDirName on the workspace stub)
```

The navigation fix is unchanged, and still untested for the reason in the commit message — the file picker doesn't open under the harness, so my first attempt passed with the guard deleted. Worth 10 seconds on staging: select a changed skill and confirm you stay in the panel.

Gate: typecheck, build, **2418 tests** and `ds:check` green.

h3xxit:
Both cubic findings addressed — the second one was a real bug and my fix genuinely did not work.

### P1 · group check never fired — **valid, fixed**

Cubic is right, and it was worse than "can still switch": it could **never not** switch. Review-session paths are **workspace-relative** — `DiffService.resolvePair` resolves them against the workspace directory, so they arrive as `knowledge-base/Groups/…` (matching the lock lines in staging's log, `path=staging-repo/roles.yaml`). `stripJunkBeforeKbDir` returns `segs.slice(idx)`, **keeping** the kb dir as the first segment, so `groupOfPath` saw `knowledge-base` where it wanted `Groups` and answered `null` for **every** path, group or not.

Now `isGroupItemPath`, which removes the `<kbDirName>/` prefix before asking — and **exported so it can be tested directly**. That matters here: the behaviour lives behind a dropdown the component harness cannot open, which is why my earlier UI-level attempt passed with the guard deleted and got removed.

11 cases: both path shapes, personal groups, the `Groups/` folder itself, and a `<kbDirName>-backup/` lookalike that neither strip may mistake for the clone (one test of mine asserted the wrong expectation there — the code was right).

**Verified they catch it:** reinstating the original predicate fails exactly the four workspace-relative cases.

### P1 · unpinned action tags in `docker-publish.yml` — **not this PR**

Correct in itself, but that file isn't mine. The branch had accidentally picked up `aaecdd1` ("Publish Docker image to GHCR; add Glama MCP directory manifest", Juan) which is **not on `dev`** — so this PR would have quietly carried someone else's commit into a merge.

Rebased onto `origin/dev`; the PR is now one commit over five files, and `docker-publish.yml` / `glama.json` are gone from the diff. The pinning question belongs wherever that commit lands, and it's a fair point there — a package-write `GITHUB_TOKEN` running mutable tags is worth pinning to SHAs.

Gate: typecheck, build, **2429 tests** and `ds:check` green.

PR #65, opened 2026-08-11
fix(library): resolve skills nested under a category folder

h3xxit:
Reported from production: every Engineering skill lists fine but reports **"This skill doesn't exist, or you don't have access to it"** when clicked. Not an access problem — the lookup never gets that far.

## Cause

A group may carry category folders. `skills.service` walks until it finds a `SKILL.md` and treats **that** folder as the skill, so `Groups/Engineering/coding/create-ticket/SKILL.md` is a skill named `create-ticket`. `WorkspaceItemRoute` read the URL as `<group>/<entry>/<rest…>`:

| URL | old `entry` | result |
|---|---|---|
| `Groups/GTM/heyreach-campaign/SKILL.md` | `heyreach-campaign` | ✅ |
| `Groups/Engineering/coding/create-ticket/SKILL.md` | **`coding`** | ❌ no such skill |

Engineering is the only group in the Bevel KB with a category layer (`coding/`, `review/`), which is why only it broke. The same assumption killed bundled files (`architecture-review`'s `check-vocab.mjs`, `pr-overview-template.html`) and any `.tool` below a category folder — `walkFiles` registers manuals at any depth, so they list and then 404.

## Fix

- **`SKILL.md` names its own skill** — the folder holding it, at any depth. Still resolved from the URL alone, so the just-created-skill case that structural resolution exists for keeps working at depth.
- **Bundled files and bare folders consult the catalog.** These are genuinely ambiguous from the URL — `<skill>/reference/x.md` and `<category>/<skill>/x.md` are the same shape — so the route takes the *deepest* skill whose folder contains the path, comparing whole segments so `create-ticket-v2` cannot claim `create-ticket`'s files. Falls back to today's reading while the catalog is loading.
- **`.tool` matches at any depth**, mirroring the backend.

Flat groups are untouched.

## Tests

Six added to the existing `WorkspaceItemRoute.test.tsx`. Verified non-vacuous: with the route reverted, exactly those six fail and all 17 pre-existing ones still pass.

Full gate green — typecheck, 2448 tests, build, `ds:check` (215, no regressions).

---
## Summary by cubic
Fix nested skill routing so skills under category folders resolve correctly. Decisions now rely on catalog facts and positive evidence, preventing 404s and false “doesn’t exist” flashes during loading or transient failures.

- **Bug Fixes**
  - The catalog is the authority for skill id and ownership; uses frontmatter id, picks the deepest owning skill, and compares whole segments to avoid prefix collisions.
  - `SKILL.md` resolves to its parent folder at any depth; a nested `SKILL.md` inside a skill is a bundled file; `Groups/<group>/SKILL.md` is supported.
  - `.tool` manuals resolve at any depth using catalog slugs, matching backend discovery.
  - Positive-evidence routing: ambiguous paths wait for the catalog; cached entries still identify categories after a failed refresh; catalog errors keep bundled-file deep links on the skill page; redirects fire only for bare category folders or files directly in the group or under a known category (never on absence).
  - Provisional names: SkillPage treats URL-derived names as provisional, defers “doesn’t exist” until the catalog has answered, and remounts on id correction to avoid one-frame error flashes.

<sup>Written for commit e7c2c4805108e9b5a46c926446bf084c556f8c30. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Bevel-Software/Hexis/pull/65?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/core-backend/package.json
+++ b/packages/core-backend/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@bevel-software/platform-core-backend",
-  "version": "0.7.3",
+  "version": "0.7.4",
   "description": "Open-source core backend of the Bevel platform: git-backed knowledge workspace, workflow (branches/change requests/locks/SSE), skills, tools, secrets vault, access control and the remote MCP surface.",
   "type": "module",
   "main": "./dist/index.js",
--- a/packages/core-frontend/package.json
+++ b/packages/core-frontend/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@bevel-software/platform-core-frontend",
-  "version": "0.7.3",
+  "version": "0.7.4",
   "description": "Open-source core UI of the Bevel platform (raw TS/TSX source): workspace explorer/viewer, branches & change requests, Library, secrets, tools, access control.",
   "type": "module",
   "main": "./src/index.ts",
--- a/packages/core-frontend/src/modules/library/__tests__/SkillPage.test.tsx
+++ b/packages/core-frontend/src/modules/library/__tests__/SkillPage.test.tsx
@@ -238,6 +238,8 @@ function renderPage(
   routerState?: Record<string, unknown>,
   /** Catalog-state overrides — e.g. `{ loading: true }` for the early case. */
   library?: Partial<LibraryContextValue>,
+  /** Props the canonical /workspace mount passes; the legacy route passes none. */
+  pageProps?: { provisional?: boolean },
 ) {
   libraryMock.value = { ...libraryValue(owned, crs, mine), ...library };
   return render(
@@ -255,7 +257,10 @@ function renderPage(
                 leave the one assertion that matters unmakeable. */}
             <LibraryToastProvider>
               <Routes>
-                <Route path="/skills-and-tools/skills/:name" element={<SkillPage />} />
+                <Route
+                  path="/skills-and-tools/skills/:name"
+                  element={<SkillPage {...(pageProps ?? {})} />}
+                />
               </Routes>
             </LibraryToastProvider>
           </EventBusContext.Provider>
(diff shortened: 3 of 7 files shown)


h3xxit:
All four addressed in the latest push — each valid, and the first one (P1) was the root of the other three: the original cut read a skill's identity off the URL wherever it could, when the catalog already knows it better.

**P1 — `SKILL.md` URLs can misroute.** Confirmed on all three counts. A skill's id is its frontmatter `id`/`name`, only *falling back* to the folder name, so a declared id resolved to the wrong name; `<skill>/examples/SKILL.md` is a bundled asset (the backend's walk stops at the first `SKILL.md`) and was being read as a skill called `examples`; and `Groups/<group>/SKILL.md` was excluded by the depth guard even though the walk would report the group folder itself as a skill.

Taken your shape: **the catalog is consulted first, for every URL**. The structural reading survives only as the fallback — which is what still lets a just-created skill open from its URL alone before any reload lands.

**P2 — the fallback names the category.** Right, and "corrects itself on the next render" was doing real work in that comment: it renders a SkillPage for a category name, which fetches, 404s, and flashes an error first. Since the catalog is the only thing that can settle the ambiguous shapes, the route now **waits** while it's loading rather than guessing. The layout and sidebar are already on screen around it, so this is a held slot, not a blank page.

**P2 — stray file under a category folder.** Widened the group-page redirect to any depth, so `Groups/Engineering/coding/access.md` gets the same answer a group's own `access.md` already got.

**P3 — the `.tool` test proved only the filename fallback.** Correct — `deploy` was absent from the fixture, so `catalogSlug` was always undefined. It now uses a manual whose declared slug (`internal_deploy`) differs from its filename, which fails if catalog refinement stops working at depth.

Four tests added, one per finding, each verified to fail against the previous commit. Full gate green: typecheck, 2452 tests, build, `ds:check` (215, no regressions).

h3xxit:
All three valid, and they share a root: an unanswered catalog is not evidence of absence. `GroupPage` already draws that line ("we couldn't ask" is not "it doesn't exist") — this route and `SkillPage` didn't.

**Bare category-folder URL renders a nonexistent skill.** Right. The discriminator you asked for exists in the data already: a category folder has catalog skills *under* it, a just-created skill whose reload hasn't landed has none. So that's the test now — `Groups/Engineering/coding` goes to its group, `Groups/Engineering/coding/brand-new-skill` still opens as a skill. Both covered.

**A transient catalog failure loses the URL.** Agreed, and this was the worst of the three: an outage would bounce every bundled-file deep link to the group page. `data.error` is now handled separately from a settled empty catalog, so those links keep their skill page — and since the detail comes from a different endpoint, it can still answer or recover on retry.

**Provisional name can still flash "doesn't exist".** Correct, and the fix belongs in `SkillPage` rather than the route: deferring in the route would have cost the catalog-free path that lets a newly created skill open at all. `SkillPage` now holds that verdict while the catalog is loading — the same guard `GroupPage` makes a few files over — so the name it was handed gets a chance to be corrected before the page calls it missing.

One thing I did **not** do, for the record: on a failed catalog, a bundled file nested deeper than its skill folder (`<skill>/reference/x.md`) still can't be attributed to the right skill — nothing in the URL distinguishes it from `<category>/<skill>/x.md`. The fallback reads the file's own folder, which is right whenever assets sit beside `SKILL.md`. The test says so rather than pretending otherwise.

Three tests added, one per finding, each verified to fail against the previous commit. Full gate green: typecheck, 2456 tests, build, `ds:check` (215, no regressions).

h3xxit:
All six addressed. They share one root, so I restructured rather than patching each: `catalogSpoke` treated the catalog as a single trust signal, when its two kinds of fact have different lifetimes.

**What the catalog knows** is trustworthy whenever it's there — `useLibraryData` keeps its entries through a failed refresh, and a category with known skills beneath it is still a category. **What it doesn't know proves nothing**: it may be loading, stale, or failed. The route no longer infers "this is not a page" from absence at all; redirects require positive evidence.

That fixes two of the findings directly:

- **Cached category after a failed refresh** (P2, conf 10) — correct, `catalogSpoke` was suppressing the very check that identifies a category. It now runs regardless of `error`.
- **Stale-catalog bundled file** (P2, conf 6) — the sharpest one. A just-created skill's `<newSkill>/notes.md` was bounced to the group page when the catalog was merely stale, which is the same symptom this PR exists to remove, via a different unsettled state. My fallback comment claimed to handle it and the code never reached that branch. A file is the group's business only when it *cannot* be a skill's: directly in the group folder, or inside a folder the catalog knows is a category.

**Provisional names** (P2, conf 9 + P3, conf 8): `SkillPage` now takes `provisional`, marking a name the route guessed from the URL rather than one the catalog resolved. Absence is a verdict a guess hasn't earned, so a failed lookup stays unconcluded until the catalog answers *successfully* — you were right that `loading: false` also covers `error`. A confirmed name is untouched and reports its error immediately, which preserves the "surface a real detail error" case you flagged. The page is also keyed by name, so correcting provisional → declared id remounts instead of rendering one frame with the old name's failed state.

**Nested bundled-file deep links under a failed catalog** (P2, conf 10): resolved as a consequence rather than by path-aware resolution — with the guard above, `.../create-sales-deck/reference/LESSONS.md` now holds an unresolved/loading state instead of asserting "doesn't exist". The name we hand it is still the file's immediate parent, because nothing in the URL can distinguish `<skill>/reference/x.md` from `<category>/<skill>/x.md`; the difference is that a wrong guess no longer produces a false verdict. The deliberate trade-off: while the catalog is genuinely down, that panel stays busy rather than resolving — honest about "we couldn't ask", and it recovers on retry.

**Test asserted only absence** (P3): fair — it would have passed on a blank render. It now asserts the busy panel is actually shown.

Six tests added or reworked; four verified to fail against the previous commit (the other two cover behaviour that was already correct and now has a guard). Full gate green: typecheck, 2461 tests, build, `ds:check` (215, no regressions).

PR #68, opened 2026-08-11
perf(pending-commits): drain in bursts, validate once per burst

h3xxit:
The "why did applying take so long?" fix. Two costs compounded on every bulk change; the second was the larger.

## What was slow

**One row per sweep, then always sleep.** [`drainOnce`](https://github.com/Bevel-Software/Hexis/blob/dev/packages/core-backend/src/modules/workflow/pending-commits.worker.ts) claimed exactly one row per workspace, and `runLoop` slept `POLL_INTERVAL_MS` after every pass whether or not it found work — capping a workspace at **~2 commits/second** however fast git was.

**Every commit re-parsed the whole KB.** `commitFile` ran the advisory commit-validation hook per file. In the enterprise app that hook is `kbValidator.runValidation`, which parses the entire knowledge base to produce a report that is only `console.warn`'d — nothing reads it, nothing writes from it. A migration touching ~75 files paid ~75 whole-KB parses to log the same 44 pre-existing issues ~75 times.

Together that's why applying a change request looked like it hung: the queue was still draining commits behind it.

## What changed

- A sweep **drains a workspace's ready rows in one pass**, bounded by `MAX_BURST_PER_WORKSPACE` (50). That ceiling is fairness, not throughput — the loop holds the single in-flight commit slot, so an unbounded drain would park every other user's save behind one workspace's migration.
- The validator runs on the **last commit of each burst**, whose tree is the end state all the per-file runs were converging on. `commitFile` already had a `skipValidator` seam (the reject flow uses it), so this is plumbing, not new machinery.
- `hasReadyRow` is a peek using the **same** readiness rule as `claimNext` — both now share one extracted predicate so they cannot disagree. It counts only *ready* rows deliberately: keying off "any pending row" would let one stuck row suppress validation on every future burst.

## Safety

A failing row cannot spin the new loop — `markTransientFailure` leaves `lastAttemptedAt` set, so the backoff gate refuses to hand the same row back within the sweep. There's a test for exactly that, plus one for `stop()` landing mid-burst.

Five tests added; four verified to fail against the previous commit (the other two guard behaviour that held trivially when only one row was drained per pass).

Full gate green: typecheck, 2466 tests, build, `ds:check` (215, no regressions).

---
## Summary by cubic
Speeds up pending commits by draining work in bursts and validating once per burst, cutting per-file polling and repeated KB parses. Also fixes `skipValidation` wiring, skips wasted peeks at the burst cap, prevents stuck queues on peek failures, and always validates the last commit at the cap.

- **Performance**
  - Drain ready rows per workspace in one sweep (max 50) for fairness.
  - Validate only the last commit of each burst; earlier commits set `skipValidation` (now passed through to `WorkflowService`).
  - Share the readiness predicate between `claimNext` and `hasReadyRow`; backoff gate prevents re-claim spin.
  - Handle peeks efficiently: failures assume “last” to avoid stranding rows, and at the burst ceiling we skip the peek and still validate the final slot.

- **Dependencies**
  - Bump `@bevel-software/platform-core-backend`, `@bevel-software/platform-core-frontend`, and `@bevel-software/platform-shared` to 0.7.5.

<sup>Written for commit 79255174654ce02a4974c6458e8ea6aa7fa3e544. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Bevel-Software/Hexis/pull/68?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/core-backend/package.json
+++ b/packages/core-backend/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@bevel-software/platform-core-backend",
-  "version": "0.7.4",
+  "version": "0.7.5",
   "description": "Open-source core backend of the Bevel platform: git-backed knowledge workspace, workflow (branches/change requests/locks/SSE), skills, tools, secrets vault, access control and the remote MCP surface.",
   "type": "module",
   "main": "./dist/index.js",
--- a/packages/core-backend/src/core/create-core-services.ts
+++ b/packages/core-backend/src/core/create-core-services.ts
@@ -622,10 +622,14 @@ export async function createCoreServices(
   );
   const pendingCommitsWorker = new PendingCommitsWorker({
     service: pendingCommitsService,
-    workflow: {
-      runPendingCommit: (workspaceId, branch, path, user) =>
-        workflowService.runPendingCommit(workspaceId, branch, path, user),
-    },
+    // The service IS the driver — passed directly rather than wrapped in a
+    // forwarding lambda. A lambda that names its parameters silently drops any
+    // the interface later adds (TypeScript accepts a shorter parameter list),
+    // and that is not hypothetical: the wrapper here took four arguments and
+    // swallowed the `opts` carrying `skipValidation`, so the burst-validation
+    // skip was a no-op in production while every unit test — which stubs this
+    // port — passed. No wrapper, no gap to fall through.
+    workflow: workflowService,
     // Both escalation sinks come through CorePorts: the recovery agent and the
     // notice sink are enterprise-owned (background-agent factory / feedback
     // dashboard); core defaults skip recovery and log to stderr.
--- a/packages/core-backend/src/modules/workflow/__tests__/pending-commits.worker.test.ts
+++ b/packages/core-backend/src/modules/workflow/__tests__/pending-commits.worker.test.ts
@@ -316,6 +316,9 @@ describe('PendingCommitsWorker.drainOnce', () => {
       expect(calls).toHaveLength(50);
       expect(calls[49]?.[4]?.skipValidation).toBe(false);
       expect(calls[48]?.[4]?.skipValidation).toBe(true);
+      // …and does not ASK on that slot: the ceiling already answers it, so
+      // peeking would be a wasted round-trip on every sweep of a backlog.
+      expect(service.hasReadyRow).toHaveBeenCalledTimes(49);
     });
 
     it('commits the claimed row even if the peek throws', async () => {
--- a/packages/core-backend/src/modules/workflow/__tests__/pending-commits.worker.test.ts
+++ b/packages/core-backend/src/modules/workflow/__tests__/pending-commits.worker.test.ts
@@ -279,19 +279,61 @@ describe('PendingCommitsWorker.drainOnce', () => {
       expect(workflow.runPendingCommit).toHaveBeenCalledTimes(1);
     });
 
-    it('does not re-claim a row that just failed — the backoff gate holds', async () => {
-      // `markTransientFailure` leaves `lastAttemptedAt` set, so `claimNext`
-      // refuses the same row until its backoff elapses. Were that not true,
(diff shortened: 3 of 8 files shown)


h3xxit:
Four fixed, one skipped with reason. The first finding was the important one — thank you for it.

**P1: the adapter drops `opts`.** Correct, and it means **half this PR did nothing**. The `workflow` port was wired through a forwarding lambda in `create-core-services.ts` that named four parameters and swallowed the fifth. TypeScript accepts a shorter parameter list, and every worker test stubs that port, so the coalescing was a no-op in production with a fully green suite — the one place my tests structurally could not look.

Fixed by deleting the lambda rather than repairing it: `WorkflowService` satisfies `WorkflowCommitDriver` directly, so there is no wrapper left to drift out of sync the next time the interface grows an argument.

**P1: a peek failure strands the claimed row.** Right, and worse than it first looks — the row is already `running` and nothing resets that, so one transient `hasReadyRow` error would stop that workspace draining until the process restarted. The peek is an optimisation, never a gate: it now falls back to "assume last", which costs one extra validation pass and nothing else.

**P2: the capped burst skips validation on its last commit.** Correct — every sweep of a backlog larger than the ceiling ended on an unvalidated commit. The final slot of the pass now validates regardless of what remains queued.

**P3: the backoff test is vacuous.** Correct, and this is the second time in this stack I've written a test that would pass with the thing it names deleted. `claimNext` is stubbed, so it proves only that the loop ends. Rewritten to state what it actually pins — a transient failure is recorded and the pass continues — and to name `readyPredicate` as where the real gate lives, noting plainly that there is no database-backed test for it in this package.

**P2: a burst ending in a no-op row can validate zero times — skipping.** The reasoning is right: `commitFile` returns before validation for a clean path, so if the terminal row is a duplicate save, that burst logs nothing. I'm leaving it because the consequence is confined to a log line — the hook is advisory, returns a report nothing reads or writes from, and no state depends on it having run. Closing it would mean parsing the whole KB on behalf of a commit that did not happen, or exposing a standalone "validate now" entry point purely to guarantee a warning. The next burst validates as normal. Happy to add it if you'd rather have the guarantee.

Three tests added, one rewritten. Full gate green: typecheck, 2468 tests, build, `ds:check` (215, no regressions).

h3xxit:
Only one of the three open threads is from the latest run; the other two are stale or already answered.

**P3 — the peek is wasted at the ceiling.** Valid, fixed. `lastOfBurst` was `!more || true` on that slot, so `hasReadyRow` ran and its answer was discarded — one pointless round-trip per sweep, on exactly the hot path this PR is about. The ceiling check now runs *before* the peek, and the peek only when its result can still decide something. Failure semantics are unchanged: a throw still means "assume last", so a claimed row is never abandoned in `running`. The ceiling test now also asserts the peek is called 49 times, not 50.

**P1 — the adapter drops `opts`.** Already fixed in `2ac9c0e`, before this run; the thread is from the 15:35 pass and just wasn't auto-resolved. `create-core-services.ts` now passes `workflowService` directly rather than through a forwarding lambda, so there is no parameter list to fall out of sync. Worth confirming on the current head rather than taking my word for it — and your newest run did not re-raise it.

**P2 — a burst ending in a no-op row can validate zero times.** Still open because I'm deliberately not fixing it, reasoning in the previous comment: the consequence is confined to an advisory log line, and closing it would mean parsing the whole KB on behalf of a commit that did not happen. Say the word if you'd rather have the guarantee than the saved parse.

Gate green: typecheck, backend 1365 passed / 2 skipped, build.

PR #93, opened 2026-08-25
Shell tools run only where they should, and agents get a declared secret allowlist

h3xxit:
The platform half of the AEL work. The knowledge-base half is change request #36; the runner half is `feat/ael-generic-runner` on coding-agent.

## `@utcp/cli`, and why the executor is deleted

A `.tool` that shells out is how a local runtime gets a git surface without an agent ever holding a token. The platform has to *understand* those files — scan, validate, list, serve the manual body to the local server that will run it — so the `cli` serializer must be registered or every such file drops out of the catalog.

What it must never do is run one. A `.tool` is knowledge-base content and agents write to the knowledge base, so anyone able to land a file there would otherwise have a shell on production.

`@utcp/cli`'s module-level `register()` installs two things: the call-template serializer and a `CliCommunicationProtocol`. Only the first is wanted, so the protocol is removed from the SDK registry immediately after the import that added it. **The hosted process therefore cannot dispatch a shell command at all** — a cli template that reached a hosted client by any route fails to dispatch rather than executing.

Forcing `remote: false` on any manual containing a cli template sits on top of that as a second, independent mechanism, and a declared `remote: true` beside one is refused rather than quietly corrected — silently fixing it would leave the catalog disagreeing with the file about what the platform will do.

Two mechanisms, because one of them being wrong is a remote shell.

## The only routes that return secret values

Everywhere else a credential is resolved in-process at call time; `resolve()` had exactly one caller. Two runtimes outside this process legitimately need values anyway — the local MCP server executing a `remote: false` tool on someone's machine, and the execution layer starting a session whose app under test has to boot.

One discipline governs both, and it is why these are two narrow routes rather than one generic "resolve these keys" endpoint: **the caller never names the variables**. It names a knowledge-base file; the server re-reads that file from the default branch and resolves exactly what it declares. The knowledge base is the allowlist, it is reviewable as a diff, and no caller — however privileged its connection key — can widen it.

Agent secrets are keyed `agent:<slug>:<VAR>`. Tool namespaces are built by mapping every non-word character to `_`, so no `.tool` can produce a key containing `:` — the two namespaces are provably disjoint in one flat keyspace rather than disjoint by naming convention. A newly landed `.agent` naming someone else's credential resolves nothing.

Read access is required (a 404, not an empty object — whether an `.agent` exists is itself information), only `remote: false` manuals qualify, responses are `no-store`, and every resolve logs the caller, the file and the variable *names*.

## hexis-mcp: local tools stop depending on the laptop

A local-only `.tool` executes on the user's machine, so its credentials had to be there too — every person running one hand-placed its secrets in their MCP client config, and a rotation had to be repeated everywhere it had been copied.

hexis-mcp now registers the `cli` protocol *with* its executor (the mirror image of the platform) and carries a UTCP variable loader that asks the deployment to resolve what a local manual's own `.tool` declares. Values are bound to that manual's UTCP namespace, so one local tool cannot reach another's secret, and nothing is returned to a caller — it is substituted into a tool invocation and goes no further. An agent calls `git.push(...)` and never sees the token.

`process.env` stays underneath as UTCP's last tier, so an existing setup keeps working; the vault answer simply arrives first when there is one.

## `http_get`

Gates poll — a health endpoint, a deployment that should be back. The only way to do that from a code-mode chain was to shell out, which runs the polling with whatever the surrounding process has in its environment and leaves the egress undeclared.

Every constraint is fixed rather than configurable, because the constraints are the feature: GET only (a tool that could POST is a request forger behind every connection key), public hosts only through the existing SSRF guard rather than a second copy of that judgement, no caller-supplied headers, and a capped body and timeout. A non-2xx comes back as a result — "not up yet" is usually the answer the gate wants, and raising it would make the ordinary case indistinguishable from a broken chain.

The plan called for this as an `http.tool`. It cannot be one: core refuses any `.tool` referencing `${API_URL}`, precisely so author-written content can never carry platform credentials.

## Also

`.pipeline` and `.agent` carry their own access verbs, which turned out to need only two entries in `ACCESS_FRONTMATTER_EXTENSIONS` — `.tool`'s whole-document convention already covered the rest.

The Secrets page grows an Agents section. A secret nobody can set is a secret nobody will use.

## Verification

Typecheck clean. 1854 backend tests, 200+ frontend, all green — including new coverage for the deleted cli executor, the namespace disjointness, the allowlist refusing to be widened by a request body, and the SSRF refusals reaching no network.

---
## Summary by cubic
Shell `.tool`s are parsed on the platform but never executed; any manual with a `cli` template is forced local, and local MCPs resolve only the variables those files declare. Manual namespaces are now unique (not just names), preventing collisions like `a-b` vs `a_b` from sharing one secret space.

- New Features
  - Registers `@utcp/cli` for parsing only and removes its executor; a manual with a `cli` template is forced `remote: false`, and `remote: true` is rejected.
  - Adds POST `/api/agent/local-tools/:slug/variables` that re-reads the `.tool` and returns only its declared variables; unreadable files 404, remote manuals 409, responses are `no-store`, and browser sessions are refused.
  - `list_local_tools` now includes each manual’s `slug` used by the variables route.
  - `hexis-mcp` registers `@utcp/cli` with its executor and adds a UTCP variable loader that resolves a local manual’s variables from the deployment; values are namespace-bound, never exposed to callers, cached per manual for a few minutes only on successful fetches, ambiguous namespaces resolve to nothing, bindings are isolated per server and released on shutdown or client-build failure, and malformed responses are treated as failures.
  - Introduces `registerAccessFrontmatterExtensions`; core defaults to `['.md', '.tool']`, and overlays can add extensions for per-file access verbs. Registration is additive and throws on malformed extensions.

- Migration
  - Upgrade `hexis-mcp` so local tools resolve secrets from the deployment instead of relying on `process.env`.
  - Author shell `.tool`s with `remote: false`; a `cli` template with `remote: true` is refused.
  - Overlays that govern `.agent`/`.pipeline` must call `registerAccessFrontmatterExtensions(['.agent', '.pipeline'])`; core’s agent env routes and Secrets UI section were removed.
  - Removed `http_get`; run probes as local `.tool`s where the work executes.

<sup>Written for commit ebba2cb159fe0d27f8863baa5d38ff771a1c7c67. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Bevel-Software/Hexis/pull/93?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


h3xxit:
Two corrections since the description above — both remove things this PR should not have had.

**`.agent` and `.pipeline` support is gone.** The execution layer is an enterprise feature and its file kinds belong with it; building them into Hexis was wrong. Removed: the `.agent` scanner, the `/agent/agents/*` env routes, the `/secrets/agents*` provisioning routes, and the Secrets page's Agents section. The `.tool` half stays — `.tool`, `remote: false` and `list_local_tools` are core concepts and hexis-mcp is a core-published package.

The access grammar survives as *grammar* rather than a file list: `['.md', '.tool']` plus `registerAccessFrontmatterExtensions`, which an overlay calls at boot. That keeps per-file governance on the files whose edits grant capability without core knowing those files exist. Additive only — removing an extension would silently drop grants already being enforced, and the declarations scan and the reference scanner have to agree on the set or a rename strands a live grant.

**`http_get` is gone too.** I added it so a gate probe could poll a health endpoint, reasoning that the probe ran on the platform and a `.tool` with a templated URL would therefore be a server-side fetch from inside our network. The premise was the mistake. A probe runs for the execution layer, in the same container as agents that already have a shell, so a GET from there grants nothing new — Razvan's read was right that it is no different from the agent shelling out to `curl`. Moving probe execution to the local MCP server (coding-agent PR #1) makes it an ordinary `remote: false` `.tool` in the knowledge base.

Worth naming plainly: as first built, that put an SSRF-guarded outbound GET on the platform reachable by every connection key, to serve one caller that did not need it there.

The isolated VM is the same at both ends, so nothing was given up by moving it.

A changeset is included — this needs a release before bevel-platform can repin and take the enterprise half.

PR #95, opened 2026-08-25
Persist git credentials into workspace clones; surface failed pushes; contain pid/thread leaks

h3xxit:
Fixes a self-hosted incident: saves reported success while nothing reached the remote for two days (136 commits stranded locally), and in the same window the app container leaked ~4,500 tasks until nothing on the host could `fork()`.

## Root cause: the boot clone never persisted its credential helper

The KB startup phase clones the default branch into the exact directory `WorkspaceService` later adopts, but passed its credential config as `git -c … clone` — the **per-invocation** form. That authenticates the clone and writes nothing into the repository it produces. `GitService.push` runs a bare `git push` expecting the clone to carry its own credentials, found none, and died with `could not read Username … No such device or address`. The setup screen's connection test kept passing throughout because it builds its own helper for its own invocation and never touches the workspace repo — so nothing the customer tried (fresh PAT, env vars, restarts, `--force-recreate`) could help: nothing re-clones a directory that is already on disk.

**Fix** (`0d3f559`): credentials go in as `clone --config`, which git persists; the snippet lives once in `kb-fs/clone-config.ts` next to the tracking config. Adopting an existing clone **re-stamps** the helper, so clones already broken by this heal at the next boot. Regression test runs real git and asserts `--local` config — a dev machine with Git Credential Manager answers a scope-less `--get` from global config, which is exactly how this stayed invisible locally.

## Visibility: failed pushes now surface (`9021784`)

The autosave path swallowed push failures entirely (deliberately — the user is mid-edit), and the release path's typed error only reaches an editor that is still mounted; the unmount-time release `.catch()`es into `console.warn`. New workspace-scoped `git-sync-failed` / `git-sync-recovered` events (recovery emitted once, on the transition — the healthy hot path stays silent) and a `GitSyncFailedBanner` beside `PullNeededBanner`: *work is saved locally, an administrator should check the server logs*, with the sanitised git error. Failure handling itself is unchanged — saves still succeed, retries and agent hand-off keep running.

## Containment: `init: true` + `pids_limit` (`2f10166`, `32e6e11`)

No init meant orphaned grandchildren (the agent shell tool SIGKILLs `sh`, not what's under it) accumulated as unreaped zombies; no `pids_limit` meant the leak spent host-wide headroom. Both set in all four shipped compose files.

## Upstream leak fixes (`08124b9`)

Bumps to the two libraries fixed for this incident:
- **`@utcp/code-mode` 1.2.13** ([code-mode#46](https://github.com/universal-tool-calling-protocol/code-mode/pull/46)) — a bridged tool call that never settled pinned an isolate OS thread in `applySyncPromise` forever (dispose's terminate interrupt needs JS to resume). Matches the customer's signature exactly: threads on one node process, flat memory.
- **`@utcp/mcp` 1.1.5** ([typescript-utcp#37](https://github.com/universal-tool-calling-protocol/typescript-utcp/pull/37), fixes #34) — dead sessions evicted on auth failures / timeouts, revoked OAuth tokens dropped, identity-aware cleanup.

## Verification

- Backend 1828 passed, frontend 1499 passed; full workspace build + typecheck clean; design-system ratchet no regressions.
- Every new regression test verified to fail against the pre-fix code.
- Three changesets ride the fixed version group.

## Still open (deliberately out of scope)

- Auth-class push failures still burn the pending-commit worker's full transient budget before escalating — with the banner in place this is no longer silent, but a faster escalation is worth its own PR.
- Enterprise repin after this releases.


h3xxit:
All ten findings were valid — addressed in b17a370, each with a test that exercises the real failure mode. Grouped:

**Banner — the two P1s were the serious ones.** The id comparison was raw while local ids are URL-encoded and SSE event ids arrive decoded, so on every slashed feature branch (`<user>/<slug>` — i.e. every draft) both guards dropped every event and the banner never showed: silently recreating the exact invisible-failure symptom this PR fixes. Both sides now canonicalize like every other SSE consumer, and the filtering tests drive the real `razvan-radulescu%2Ffeat` vs `razvan-radulescu/feat` split instead of slash-free ids. The two stale-banner findings (same root cause, two reports): the failure state now remembers which workspace it belongs to, gated at render — not an effect reset; `react-hooks/set-state-in-effect` flags that, and the ownership gate is also better behavior: hidden off-workspace, still shown on returning to the genuinely-failing branch.

**Credential rotation (P1).** Right — a token added or username changed through the setup screen after a branch was cached never reached that clone until a restart. The cached fast paths now re-stamp when a credential *fingerprint* changes: username + token-presence only, deliberately excluding the token value, because the helper reads `$GITHUB_TOKEN` at call time — plain rotation needs no re-stamp, so the check stays a string compare on the hot path. The lost-token case now unsets the clone-local helper on adoption (tolerated no-op when absent), so a stale helper can't answer with an empty password over whatever auth the operator fell back to.

**Workflow pushes (P2s).** The five direct pushes (share, CR-open, update-from-base, decline revert, roles restore) now route through a `trackedPush` that notes failure *and* success — so a failed share raises the banner and a successful one clears it — while preserving the exact bare-push call shape (an explicit `undefined` third arg is a different signature to every spy). And when cooperative recovery ran and failed, the banner now carries the recovery error rather than the first rejection the pull may already have cured.

**The two P3 test findings** were fair: the "collapse" assertion only ever saw a fresh clone, and the "independence" test never left the failing state. The collapse test now drifts a second helper value in out-of-band and adopts through `WorkspaceService`; the independence test recovers ws-1 mid-sequence and asserts it neither clears nor suppresses ws-2's failure.

Backend 1831 / frontend 1501 passing, design-system ratchet clean.

h3xxit:
All four round-two findings were valid — addressed in df21607, each with a test that fails without the fix.

1. **Tracker keys (P2)** — canonicalized, and the emitted event ids with them, so HTTP-triggered (decoded) and worker-triggered (encoded) pushes agree on one workspace. Two supporting facts checked rather than assumed: the SSE dispatcher already canonicalizes its focus comparison, so decoded emits still deliver; and `assertValidBranchName` bans `%`, so the decode is idempotent for every legal id.

2. **Fingerprint on failed stamp (P1)** — the fingerprint is recorded only after a clean stamp (the tolerated unset-of-a-missing-value counts as clean, since the clone is genuinely in the desired state) and dropped on failure, so the cached fast path retries until a write lands. A concurrent `.git/config` lock collision therefore degrades to one warning plus a retry on the next open.

3. **`--unset-all` scope (P2)** — both the unset *and* the `--replace-all` (which had the same clobbering behavior toward operator helpers, though unflagged) are now scoped by a value pattern matching only app-stamped helpers (`password=$GITHUB_TOKEN` — a literal no operator-authored helper has a reason to contain). New integration test: an operator's `cache --timeout=300` survives our stamp and our token-removal unset; the drift-collapse test now drifts an app-shaped duplicate, which is the accumulation that actually occurs.

4. **Per-branch banner records (P2)** — one record per canonical workspace; branch B failing no longer hides branch A's unresolved failure on return, and A recovering clears only A. On the second half of the finding: with focus-scoped SSE there is no event source that can refresh a non-focused branch's status, so "refresh on focus" needs a status query endpoint — out of scope here. Between the two remaining options the comment now documents choosing retention over clear-on-switch: a possibly-stale banner self-corrects on that branch's next push, while clearing hides real broken state, which is the failure mode this PR exists to eliminate.

Backend 1833 / frontend 1502 passing, typecheck, lint and ratchet clean.

PR #97, opened 2026-08-25
Dev

h3xxit:


h3xxit:
Three of the four findings were valid and are fixed in 54b2066 on dev; the P1 is refuted with a live demo.

**P1 (value pattern "over-escaped") — not valid.** The claim reads the *source* escaping as the runtime string: the TS literal `'password=\$GITHUB_TOKEN'` yields backslash-dollar at runtime, which is a POSIX ERE literal `$` — not a backslash plus an end-anchor. Ran it against real git to settle it: seeded a clone with an app-shaped stale helper plus `cache --timeout=300`, then `--replace-all` with the pattern → the app-shaped value is replaced in place (no duplicate appended) and the operator helpers are untouched; `--unset-all` with the pattern → exit 0, only ours removed. The PR's real-git integration tests (drift collapse to one value, operator-helper survival, token-removal unset) also could not pass if the pattern matched nothing — they run actual `git config` against actual clones.

**`echo` → `printf` (P2) — valid.** POSIX leaves echo's backslash handling implementation-defined (dash treats `\c` as stop-output), so a token with backslash sequences could be silently truncated. The snippet now uses `printf '%s\n'`, verified by executing the generated helper under `sh` with a `\c`-bearing token — it passes through verbatim. The `password=$GITHUB_TOKEN` literal stays in the snippet, so echo-era helpers stamped by earlier builds still match the value pattern and get replaced on the next stamp — the migration is free.

**Throw containment (P2) — valid.** The username-validation throw could escape into `getOrCreateForBranch`'s adoption catch, which reads every error as "not on disk" and re-clones into the same validation failure — masking the cause. `stampCredentialHelper` now contains it as a loud refusal (fingerprint dropped so it retries), restoring `normalizeCloneConfig`'s never-throws contract.

**Autosave banner gating (P2) — valid, and the right call.** A background autosave losing a non-fast-forward race is the routine two-editors condition; the release push retries it and raises the banner itself (with the agent hand-off) if it truly fails. The autosave path now banners only the class that cannot clear itself — auth/host failures — which is exactly the original incident. Test updated: NFF-with-failed-recovery asserts *no* banner; the auth-failure banner assertion is unchanged.

Backend suite 1833 passing.

PR #100, opened 2026-08-26
Knowledge's empty state offers recently changed pages instead of every folder's access.md

edujuan:
## The problem

With nothing open, Knowledge offers the reader somewhere to start. On a real
knowledge base every one of those offers was an `access.md`.

The offer was built by walking the file tree breadth-first and taking the
shallowest markdown it found. Every folder carries the `access.md` that governs
it, and that file sits one level *above* the pages inside, so the walk reached
every access file in the tree before it reached a single page. The walk did
skip an `access.md`, but only the copy at the repository root, and it skipped it
by starting below the root rather than by recognising the name.

That makes it structural, not an edge case: there is no knowledge base with
subfolders where this self-corrects. The front door of the product offered four
permission files.

## What changed

**The offer now comes from the branch's history** — the most recently changed
pages this reader is allowed to open, newest first. That is the question a
knowledge base should answer for someone arriving at it, and only the history
can answer it: the file tree carries no timestamps, and position in the tree
says nothing about what the team is working on.

- `GitService.recentlyChangedPaths` walks the log newest-first, drops deletions,
  and intersects what remains with the index (a file edited in one commit and
  deleted in a later one is still named by the earlier one). The index is listed
  whole rather than passed as a pathspec: several hundred long KB paths on argv
  would blow Windows' ~32K command-line limit, and `ls-files` is the one command
  in that file which cannot read pathspecs from stdin.
- `GET /workspace/:id/recent-pages` serves it through three filters: shape
  (markdown, never an `access.md`), scope (the area the explorer browses under
  Knowledge), and read — `buildTreeReadFilter`, **the same verdict the file tree
  itself is built with**. Recency can never widen what a reader sees.
- The tree walk stays as the fallback for a branch with nothing committed yet,
  and it no longer offers an access file at any depth.

**`isAccessMdPath` moves to `kb-layout.ts`** so both sides of the app ask one
question and get one answer. The access model re-exports it, so the resolver and
the approval gate are unchanged.

## Verification

- Frontend: 1508 tests pass (122 files)
- Backend: 1150 tests pass across the touched modules
- Typecheck clean across all 7 packages; commits 2 and 3 typecheck independently,
  so the history bisects
- Lint: no new problems on any changed file
- New coverage: the tree walk at every depth, the git recency walk (including a
  non-ASCII filename and the edited-then-deleted case), and seven route tests —
  the read gate, scope, access-file exclusion, limits, and the failure path

## Notes for reviewers

- **`platform-shared` takes a minor, not a patch.** `IWorkflowService` gains a
  required `listRecentlyChangedPaths` and the interface is published, so an
  outside implementer has to add the method.
- **Two findings were fixed during self-review before this was opened.** The
  existence check originally passed candidates as an argv pathspec (Windows
  command-line limit) and swallowed its own failure, which would have left the
  feature silently dead there. It now lists the index and lets the error reach
  the route, which logs and falls back.
- **Pre-existing, not from this branch:** `@utcp/cli` is declared in
  `packages/core-backend/package.json` but absent from the lockfile's installed
  set, so `steps.test.ts` cannot load. Reproduced on pristine `dev`. `pnpm
  install` is currently blocked by the `minimumReleaseAge` supply-chain policy
  and an `overrides` lockfile mismatch; I did not rewrite the lockfile.

---

---
## Summary by cubic
Knowledge's empty state now offers the most recently changed pages the reader can open, instead of `access.md` permission files on every knowledge base with subfolders.

- New `GET /workspace/:id/recent-pages` serves candidates from the branch's git history through the same read filter the file tree uses, so recency never widens what a reader sees.
- The tree walk stays as a fallback for branches with no commits and now skips `access.md` at any depth.
- A failed history read logs a warning and returns empty, leaving the frontend its tree-walk fallback.
- `isAccessMdPath` moves to `kb-layout.ts` so both sides of the app share one definition.

**Migration**
- `platform-shared` takes a minor release: `IWorkflowService` gains a required `listRecentlyChangedPaths` method, which external implementers must add.

<sup>Written for commit f42f321746dfbb1e89992e8497ddf91782a5e8d8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Bevel-Software/Hexis/pull/100?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


edujuan:
/bevel-review

Code-level pass is clean (cubic: no issues across 15 files, zero inline threads).
Requesting the architecture pass.

Shape of the change, for orientation:

- A new read surface, `GET /workspace/:id/recent-pages`, answering "where should
  I start reading?" from the branch's git history rather than from tree position.
- `GitService.recentlyChangedPaths` as the recency primitive, plumbed through
  `IWorkflowService` (a required addition to a published interface, hence the
  minor bump on `platform-shared`).
- The route's access story is the part most worth scrutinising: history knows
  nothing about who is asking, so candidates pass through `buildTreeReadFilter`,
  the same verdict the file tree is built with, on the same batched
  `canReadBatch` path. The intent is that recency can never widen what a reader
  sees.
- `isAccessMdPath` relocated to `kb-layout.ts` as the single source of truth for
  "what counts as an access file", with the access model re-exporting it.

---

edujuan:
/bevel-review

Re-triggering: the previous run (32985727801) never got a runner during
yesterday's Actions outage, so it failed with its `start` job still queued and
posted no status.

The branch has since been merged with `origin/dev` (now at v0.13.0, 26 commits
ahead of where this was cut). That merge touched two files this PR also changes,
`git.service.ts` and `shared/src/workflow/interface.ts`, and it landed without
conflicts. Re-verified on the merged head f42f321: all 7 packages typecheck,
ds:check shows no regressions, 1501 frontend and 1165 backend tests pass.

The architecture question is unchanged from the first request: a new read
surface answering "where should I start reading?" from git history, whose access
story leans on `buildTreeReadFilter` so recency can never widen what a reader
can see.

---

edujuan:
**Architecture review: go-ahead** (`bevel/architecture-review` = success, description `pass`).

Recording it here because the workflow could not post its own verdict comment.
Its `review` job succeeded and published the status and the one-pager; the
`publish` job's "Post the review comment" step then failed:

```
gh: Resource not accessible by integration (HTTP 403)
POST /repos/.../issues/100/comments
```

That is a token-scope problem in the workflow, not a review outcome. The job
declares `issues: write` and `pull-requests: write`, so the cap is coming from
the repository or org "Workflow permissions" setting rather than from the
workflow file. Worth fixing centrally, since it will silently swallow the
verdict comment on every PR while leaving the status green: run 33055513910
shows as failed even though the review itself passed.

One-pager: https://knowledge.bevel.software/workspace/main/knowledge-base/Data/Engineering/Knowledge/Hexis/PR-Overviews/PR-100.html

Gate summary on `f42f321`:

| Gate | Result |
| --- | --- |
| `test` (build, typecheck, tests, ds:check, notices:check) | success |
| `cubic · AI code reviewer` | success, no issues across 15 files |
| `bevel/architecture-review` | success (`pass`) |
| Unresolved review threads | 0 |
| Mergeability | MERGEABLE / CLEAN |

---

edujuan:
@h3xxit these is to fix the bug that showed the suggestions of files to open always as "access.md" --> changed the suggestions to the latest edited pages, instead of the "shallowest files in the folders" (which lead to a lot of access.md suggestions)

PR #128, opened 2026-08-31
Dev

h3xxit:
## Summary by cubic
Tool connection status now separates stored credentials from working connections: **Connected** requires a successful live probe, while untested credentials no longer claim health. The PR also hardens branch lifecycle and history access, restores local CLI tool registration, and improves deployment diagnostics. All packages bump to `0.14.0`.

**Connection health**

- Saving credentials and **Test connection** follow the real authenticated request path across OAuth, inline, HTTP, MCP, environment variables, and declared health checks.
- Untested values show **Key saved**, **Signed in**, or **No key needed**, while definitive provider rejections show **Not working** with redacted details.
- Probes enforce SSRF checks, timeouts, bounded response reads, and newest-result handling; any credential write — delete included — orphans in-flight probes and clears their verdicts and errors.
- Unsupported `sse` MCP servers are skipped and cannot be newly saved; existing entries can be edited to `streamable-http`.
- Local `.tool` references register `cli` tools without widening explicit or remote protocol restrictions.

**Security and operations**

- Branch deletion and retirement protect open requests using either branch, with atomic lifecycle locking and strict fresh-fetch checks that leave requests open when freshness cannot be proven.
- History and diff routes enforce per-file `read:` access, reject unscoped or directory paths, authorize every comparison ref, match touched names untrimmed, and fail closed while keeping each error's own status.
- Encryption-key errors name the variable that supplied the key, and `tini -s` prevents PID-1 warnings under Compose init.

<sup>Written for commit a0c613ee4a0156a1b021e59bf24034fbc5e97c03. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Bevel-Software/Hexis/pull/128?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


h3xxit:
@cubic-dev-ai review this PR

h3xxit:
The P1 on `manuals.ts` is taken on #131 (the branch that introduced it, still open): the widening now requires the literal `${API_URL}` origin — author-unreachable, since core refuses any author-written `.tool` referencing `${API_URL}` — with a negative test for the foreign-origin-with-platform-path shape. The remaining findings are on other dev changes and stay open here.

h3xxit:
All remaining findings are addressed on dev in `8b9401a` (this PR picks it up): the deleted-branch sweep now demands a provably fresh branch list — `listBranches` gains `strictFetch`, a failed fetch throws and the sweep closes nothing — so stale refs can no longer close a live request (the 1967/2015 findings); deleting a request with an unresolvable base answers 409 with the real condition instead of the admin accusation (1821); the probe scrubs URLs before reading status digits (connection-probe); the facade db stub answers from the where-clause and the late-probe test flushes the render it asserts about (both test findings). Full suites green: backend 2020, frontend 1576.

h3xxit:
Second and final batch on dev in `950a07c` + `2d2d9a2` — every remaining finding is addressed: the history gate refuses the repo-relative path form (the git layer reads both spellings as one object) and refuses directory pathspecs at the git layer (`history is served per file`, with real-git tests); compare-file authorizes at both served refs; deleteChangeRequest requires a provably fresh `origin/<base>` before reading roles.yaml there (409 otherwise); an auto-discovered OAuth bearer now reaches a declared healthCheck; sse transport survives the template rebuild; inline probes inherit no top-level headers; probe variables fall back to process.env like UTCP's own tiers; the connection section is keyed per tool, derives `checking` from the probed revision, and clears a stale transport banner on the next clean probe; the noteWorkspaceFetched flag counts a fresh clone as a successful fetch; the where-clause evaluator defaults unparseable nodes to not-a-match; and the two doc comments say what the code now does. On the on-demand-sweep finding (`workflow.service:2027`): with strictFetch the sweep can no longer treat a failed probe as completed — it aborts closing nothing and the rows stay open for the next kick; the kicking verbs already surface their own error, and coupling their responses to a background fetch is the trade the detached design deliberately refuses. Full suites green: backend 2023, frontend 1576.

h3xxit:
Round three, on dev in `e0a8934`: the tree guard checks both sides of the commit, covers /changes, and rethrows any cat-file failure that is not proven absence; compare-file authorizes the local candidates as well as origin (deny on any resolvable denial, deny when nothing resolves); sse mcp.json servers are refused with a message naming the fix — the pinned @utcp/mcp schema is stdio|http, so the transport field came back out; the probe env fallback reads only the SDK's single effectiveKey spelling; and the probe transport error is now the section's own revision-stamped banner, so no page-level callback can outlive its subject (latestRev is gone with it). Suites green: backend 2024, frontend 1576.

h3xxit:
Round four, on dev in `fe4ea37`: the mcp.json editor refuses to WRITE an sse server (naming the fix; existing entries stay readable for editing to streamable-http); compare-file fails closed on an access-model error instead of reading it as an unresolvable ref; logForFile proves per-file from what the log itself touched (--name-only, any foreign path refuses — covers a directory deleted before HEAD, with a real-git test); and the probe alert clears on every credential write and on probe start, rendered full-width below the header row. Suites green: backend 2025, frontend 1576.

h3xxit:
Round five, on dev in `04645e5`: compare-file's catch preserves the error's own status mapping (503 for an unreadable access tree — still fail-closed); the log guard compares touched names untrimmed so whitespace-edged filenames aren't rejected as directories; a credential write orphans a pending probe (sequence bump + slot release — a delete starts no replacement, so nothing else would silence it); and the handler docs sit above their handlers. Suites green: backend 2025, frontend 1576.

h3xxit:
Taken on dev in `ccbc065`: `changed()` clears the verdict too — everything a probe ever published dies with the credential it described. Frontend suite green (1576).

PR #130, opened 2026-09-01
hx-group-member-email-autocomplete — Suggest people by email when adding members to a group

h3xxit:
Ticket: [Suggest people by email when adding members to a group](https://knowledge.bevel.software/hx-group-member-email-autocomplete) (`Data/Engineering/Knowledge/Hexis/Tickets/Group-Member-Email-Autocomplete.md`)

## Summary
Adding a member to a manual group was a bare email field. The roles page next to it already suggests people (name and email) from the deployment, current members excluded. This moves that behaviour into a shared `AddMemberInput` component that both the groups page and the roles page render:

- Typing 2+ characters shows suggestions (name + email) from the same `/access/suggest` source the roles page uses, with the target's current members excluded.
- Choosing a suggestion submits through the exact same path as typing the full email.
- A free-typed email not in the suggestion list stays addable; a failed suggest request never blocks adding a valid email, it just degrades to a plain input.
- Below 2 characters, no suggestion request is made.
- The roles page's add-member input behaves exactly as before (optimistic chips, "use the group assignment instead" refusal); the groups page keeps its own `group:`-prefix refusal hint. Both are now driven by the one shared component.
- Synced (IdP) groups still render no member input at all.
- `initials` moved to `lib/email.ts` so both callers can share it.

## Testing
- `pnpm typecheck && pnpm test` — all projects clean, full suite passing (308 test files / 3802 tests), including 5 new suites for this ticket covering: suggestions appear and exclude current members, choosing a suggestion equals typing the email, a free-typed email stays addable, suggestion failure degrades to the plain input, and no request below 2 characters.
- Verified live against a from-source Docker boot (Playwright/Chromium) — see the ticket's Local Testing Evidence section for full detail.

---
## Summary by cubic
Adds people suggestions to the group add-member input, so a member can now be picked by name instead of typing a full email. Choosing a suggestion submits the same request as typing the email, and a free-typed email not in the suggestions stays addable.

**Refactors**
- Extracts the roles page's suggest-as-you-type logic into a shared `AddMemberInput` component used by the groups page and roles page.
- The roles page's input behavior is unchanged; synced (IdP) groups still render no member input.
- Suggestions are keyboard-reachable, hidden while an add is in flight, and cleared the moment the query changes; a stale response never repopulates the list over a newer query.
- A failed suggest request degrades to a plain email input; no request is made below two characters, and current members are excluded.

<sup>Written for commit d87c28c094799e02fc10a50b399de26eed0e9b39. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Bevel-Software/Hexis/pull/130?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/core-frontend/src/lib/email.ts
+++ b/packages/core-frontend/src/lib/email.ts
@@ -23,3 +23,18 @@ export const GROUP_MEMBER_PREFIX = 'group:';
 export function isGroupPrefixed(value: string): boolean {
   return value.trim().toLowerCase().startsWith(GROUP_MEMBER_PREFIX);
 }
+
+/**
+ * Avatar letters for a person, from their email or display name: two initials
+ * when the local part looks like `first.last`, otherwise the first two
+ * characters. Lives beside the email rule because every caller is an
+ * email-keyed people surface.
+ */
+export function initials(value: string): string {
+  const cleaned = value.trim();
+  if (!cleaned) return '?';
+  const [name] = cleaned.split('@');
+  const parts = name.split(/[.\s_-]+/).filter(Boolean);
+  if (parts.length >= 2) return (parts[0][0] + parts[1][0]).toUpperCase();
+  return name.slice(0, 2).toUpperCase();
+}
--- a/packages/core-frontend/src/modules/admin/__tests__/AddMemberInput.test.tsx
+++ b/packages/core-frontend/src/modules/admin/__tests__/AddMemberInput.test.tsx
@@ -2,7 +2,7 @@ import { describe, it, expect, vi, beforeEach } from 'vitest';
 import { useState } from 'react';
 import { render, screen, waitFor } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
-import { AddMemberInput } from '../components/AddMemberInput';
+import { AddMemberInput, SUGGEST_DEBOUNCE_MS } from '../components/AddMemberInput';
 import { suggestPrincipals, type SuggestResponse } from '../../access/api';
 
 vi.mock('../../access/api', async (importOriginal) => {
@@ -10,8 +10,12 @@ vi.mock('../../access/api', async (importOriginal) => {
   return { ...actual, suggestPrincipals: vi.fn() };
 });
 
-/** The component's debounce, plus slack — how long "no request" has to hold. */
-const PAST_DEBOUNCE_MS = 350;
+/**
+ * How long "no request" has to hold: the component's own debounce plus slack,
+ * derived from the component so raising the debounce cannot leave this test
+ * asserting over a window the request has not been scheduled in yet.
+ */
+const PAST_DEBOUNCE_MS = SUGGEST_DEBOUNCE_MS + 150;
 
 const ALICE = { name: 'Alice Green', email: 'alice@example.com' };
 const PAT = { name: 'Pat Kim', email: 'pat@example.com' };
@@ -24,9 +28,11 @@ function people(...list: { name: string; email: string }[]): SuggestResponse {
 function Harness({
   onSubmit = () => {},
   exclude = [],
+  busy = false,
 }: {
   onSubmit?: (value: string) => void;
   exclude?: string[];
+  busy?: boolean;
 }) {
   const [value, setValue] = useState('');
   return (
@@ -36,6 +42,7 @@ function Harness({
       onSubmit={onSubmit}
       exclude={exclude}
       inputLabel="Member email"
+      busy={busy}
     />
   );
 }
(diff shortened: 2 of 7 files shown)


h3xxit:
## Answering the review

All six findings are addressed on the new head (`4dbb60f`). Five accepted, one accepted in part — detail below.

### `AddMemberInput.tsx` L140 + L145 — rows stay live while `busy` (P2 ×2)

**Accepted, one fix for both.** The input and Add button went inert during a mutation but the rows did not, which left a live row as the one remaining way to queue a second add behind the first. Took the suggested edit: the list is now behind `!busy && showSuggest && …`, so the rows are gone rather than merely disabled. Hiding rather than disabling is deliberate — a disabled row that is still painted reads as "this person can't be added", which is not what is true.

### `AddMemberInput.tsx` L95 — stale suggestions survive the new query (P2 + P3, same root cause)

**Accepted.** `setSuggestions([])` now runs the moment a new qualifying query starts, before the debounce timer is armed. The previous behaviour left the old query's people clickable through the debounce *and* the request, so a click could add someone the current text never named.

### `AddMemberInput.tsx` L147 — keyboard cannot select a suggestion (P2)

**Accepted, and it needed more than the suggested one-line swap.** Moving the submit to `onClick` is necessary but not sufficient: the input's `onBlur={() => setTimeout(…, 150)}` tore the list out of the DOM 150ms after focus left the input, so Tab could never *land* on a row to press Enter on it. Three changes together:

- `onMouseDown` now only `preventDefault()`s (keeping focus in the input so a mouse click never blurs the widget out from under itself);
- the submit hangs off `onClick`, which both a pointer and Enter/Space on a focused row raise;
- the close decision moved from the input to the widget wrapper, which closes the list only when `relatedTarget` is outside the widget. Focus moving from the input into the list is movement *within* the widget, so the list survives it.

The 150ms blur timeout is gone with it — with mousedown prevented there is no blur left to race.

Covered by a new test that types, Tabs, asserts the row has focus, and presses Enter.

### `DirectoryGroupsPage.suggest.test.tsx` L20 — `PAST_DEBOUNCE_MS = 350` (P3)

**Accepted on the constant, disagreed on dropping the wait.**

The drift risk is real and is fixed the way you suggested: `SUGGEST_DEBOUNCE_MS` is exported and both suites now compute `PAST_DEBOUNCE_MS = SUGGEST_DEBOUNCE_MS + 150`. Raising the component debounce can no longer leave a test asserting inside a window the request has not been scheduled in yet.

I kept the wait rather than asserting immediately. The claim under test is not "no request has been *sent* yet" — it is "no request was ever *scheduled*". An immediate assertion passes even against a component that armed a `setTimeout` and would fire 200ms later, which is exactly the regression the test exists to catch. Outlasting the debounce is what makes the assertion mean something. The cost is one real 350ms wait in two suites, which seems a fair price for the only thing that proves the short-circuit.

### Tests added

Three new cases in `AddMemberInput.test.tsx`, one per accepted fix: a keyboard alone reaches a row and chooses it; no rows are offered while `busy`; the previous query's people drop the moment the text changes.

`pnpm typecheck` clean across all projects; the full `core-frontend` suite passes (127 files / 1591 tests).

---

<sub>Note on the commit shape: this environment holds no git push credential, so the three files were published through the GitHub REST API as three commits rather than one. Same content, tested together.</sub>

h3xxit:
## Answering round 2

One finding on `4dbb60f` — `AddMemberInput.test.tsx:159`, the stale-query test sending `alpat` rather than `pat`. Fixed on the new head (`d87c28c`), though not the way the comment prescribed. Both halves are worth separating.

### The observation is correct

`userEvent.type` appends, so typing `'pat'` onto `'al'` asked for `alpat`. The test named a query it never made. That's a real defect and it's fixed.

### The prescribed remedy would have removed the test's teeth

Clearing the input first takes the value to `''`, which is below the two-character threshold — and the below-threshold branch already empties the list on its own, independently of the fix this test exists to guard. Routing the assertion through that branch means the test passes whether or not the new-query clear is present.

I checked rather than argued from reading. With `setSuggestions([])` deleted from the component's new-qualifying-query path:

| test shape | fix present | fix deleted |
|---|---|---|
| replace in one edit (**this PR**) | pass | **fail** ✅ catches it |
| `clear()` then `type('pat')` (prescribed) | pass | **pass** ❌ catches nothing |

So the prescribed version would have been a test that cannot fail for the reason it exists.

### What it does instead

The text is **replaced** in one edit — focus, select-all, `userEvent.paste('pat')`. That makes the query genuinely `pat`, satisfying the finding, while never dipping below two characters, so the only thing that can drop Alice's row is the new-query clear. `expect(input).toHaveValue('pat')` pins the query down in the test itself, so this can't silently regress to appending again.

The mock is now keyed by query rather than by call order: how many requests a burst of typing produces depends on where the debounce falls, which this test has no opinion about.

`pnpm typecheck` clean; full `core-frontend` suite green (127 files / 1591 tests).

PR #132, opened 2026-09-02
Encrypt PII columns at the application layer

h3xxit:
## What

Personal data in the database is now AES-256-GCM ciphertext: emails, display names, avatar URLs (including the denormalized author/approver/holder copies on `pr_file_approvals`, `pr_merge_log`, `pr_comments`, `change_requests`, `pending_commits`, `file_locks`), change-request titles/bodies, review comment bodies, and merge/queue error text. A leaked dump, an injected query, or a compromised DB credential yields no personal data — the key lives only in the app environment. Disk encryption still carries the blanket at-rest claim; this is the DB-specific control on top of it (DPA work).

## How

- **`shared/column-crypto.ts`** — two subkeys HKDF-derived from the existing `SECRETS_ENC_KEY` (no new env var): an `encryptedText` drizzle column type (randomized ciphertext on write, transparent decrypt on read, plaintext fallback for pre-upgrade rows) and a deterministic HMAC `blindIndex` for equality.
- **Blind-index companions** — randomized ciphertext can't back `eq()` or unique constraints, so six `*_bidx` columns carry them: login lookup, user upserts (conflict target moves from `users.email` to `users.email_bidx`), bot ensures, approval insert/revoke idempotency, and GDPR erasure — which also re-keys the bidx to the placeholder so nothing linkable to the erased address survives.
- **Migration 0005 + boot backfill** — the SQL migration only adds nullable columns; `runPiiEncryptionBackfill` (runs right after core migrations on every boot, idempotent) seals existing plaintext rows, fills the blind indexes, then swaps the unique constraints (`users_email_unique` → `users_email_bidx_unq`, `pr_file_approvals_unq` → `pr_file_approvals_bidx_unq`; the new index goes up before the old drops).
- **Init** — `initColumnCrypto` runs in `CoreConfig`'s constructor (and the vitest setup), so keys exist before anything touches the database.
- Everything is exported from the package (`encryptedText`, `blindIndex`, `runPiiEncryptionBackfill`, …) so the enterprise overlay can seal its own schema's columns the same way in a follow-up.

Deliberately plaintext: branch/path/workspace columns (SQL keys, and duplicated as plaintext git refs/URLs/logs on the same host — disk encryption covers them), hashes, statuses, non-PII jsonb.

## Ops note

From this version, losing `SECRETS_ENC_KEY` loses the PII columns, not just stored credentials — UPGRADING.md and `.env.example` now say to escrow the key outside the server.

## Verification

- Typecheck clean; full suite green (2033 tests, incl. new `column-crypto` unit tests).
- End-to-end against a real Postgres 18: fresh migrate → seeded pre-upgrade plaintext rows → backfill seals them (raw SQL shows only ciphertext, `email_bidx = blindIndex(email)`) → constraint swap verified in `pg_catalog` → login/upsert/erasure through blind indexes → second backfill pass is a byte-identical no-op.

---
## Summary by cubic
Encrypts PII columns at the application layer — emails, display names, avatar URLs (including denormalized author/approver/holder copies), change-request titles/bodies, review comment bodies, and merge/queue error text are now AES-256-GCM ciphertext in Postgres. A leaked dump, injected query, or compromised DB credential no longer yields personal data; disk encryption still covers the blanket at-rest claim. Also fixes a CI flake in the SkillPage load test by awaiting the integrations section.

**Migration**
- Keys are HKDF-derived from `SECRETS_ENC_KEY` with no new env var; losing that key now loses these columns too, so escrow a copy outside the server.
- First boot seals existing plaintext rows, collapses near-duplicate approvals, and swaps unique constraints onto the blind-index columns; upgrade with a single app instance.
- Ad-hoc SQL equality against these columns (looking up a user by email in `psql`) no longer works — use the app or API.

**How it works**
- Randomized ciphertext can't back `eq()` or unique constraints, so six `*_bidx` HMAC blind indexes carry login lookups, user upserts, bot ensures, approval idempotency, and GDPR erasure.
- The boot backfill runs in one transaction under an advisory lock; a strict sealed-row check catches plaintext that merely resembles ciphertext, and compare-and-swap updates mean concurrent or mixed-version boots may delay sealing but never overwrite newer data or pair a blind index with a value it wasn't read with.
- GDPR erasure fails closed if the stored email can't be decrypted with the configured key, then re-keys the blind index to the placeholder.

<sup>Written for commit dcaee86eaf25fa61b2cb8a1345a3a1608864a61e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Bevel-Software/Hexis/pull/132?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -32,8 +32,11 @@ ADMIN_PASSWORD=
 # Signs login sessions + OAuth state. Any long random string.
 JWT_SECRET=
 
-# Core secrets encryption: 32-byte key (base64 or hex) encrypting the secrets
-# vault values + MCP OAuth client secrets/tokens.
+# Core at-rest encryption: 32-byte key (base64 or hex) encrypting the secrets
+# vault values, MCP OAuth client secrets/tokens, and (via derived keys) the
+# PII columns in the database — emails, display names, change-request/review
+# text. Keep a copy outside the server: without it, backups of that data
+# cannot be read back.
 SECRETS_ENC_KEY=
 
 # Which release docker-compose runs — the image CI publishes per release
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -31,7 +31,9 @@ From 0.14, personal data in the database — emails, display names, and
 change-request/review text — is encrypted with a key derived from
 `SECRETS_ENC_KEY`, in addition to the secrets vault it already sealed. The
 first boot after the upgrade rewrites existing rows automatically; nothing
-to do.
+to do. Upgrade with a single app instance (the normal
+`docker compose pull app && up -d` flow already replaces the container) —
+don't run old and new versions against the same database side by side.
 
 What it changes for operations:
 
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -25,6 +25,24 @@ after an upgrade can take a little longer.
 older app cannot read a newer database. To go back, restore the backup you
 took before upgrading.
 
+### PII encryption (0.14+)
+
+From 0.14, personal data in the database — emails, display names, and
+change-request/review text — is encrypted with a key derived from
+`SECRETS_ENC_KEY`, in addition to the secrets vault it already sealed. The
+first boot after the upgrade rewrites existing rows automatically; nothing
+to do.
+
+What it changes for operations:
+
+- **Losing `SECRETS_ENC_KEY` now loses this data too**, not just stored
+  credentials. Keep a copy of the key somewhere safe *outside* the server —
+  a database backup can only be read back with the key that was in `.env`
+  when it was taken.
+- Database dumps contain ciphertext for these columns; ad-hoc SQL against
+  them (looking up a user by email in `psql`, say) no longer works — use
+  the app or its API instead.
+
 ## Backups
 
 Everything that matters lives in Postgres and the named Docker volumes; the
--- a/packages/core-backend/migrations/0005_pii_encryption.sql
+++ b/packages/core-backend/migrations/0005_pii_encryption.sql
@@ -1,0 +1,14 @@
+-- PII column encryption: add the blind-index companion columns, NULLABLE and
+-- unconstrained. The columns are populated — and existing plaintext PII rows
+-- rewritten to AES-256-GCM ciphertext — by the programmatic backfill that runs
+-- right after the SQL migrations on every boot (`runPiiEncryptionBackfill` in
+-- migrate.ts). That same step then applies SET NOT NULL and swaps the unique
+-- constraints (`users_email_unique` → `users_email_bidx_unq`,
+-- `pr_file_approvals_unq` → `pr_file_approvals_bidx_unq`), because a unique
+-- index on a blind-index column can only go up once every row has one.
+ALTER TABLE "users" ADD COLUMN IF NOT EXISTS "email_bidx" text;--> statement-breakpoint
+ALTER TABLE "pr_file_approvals" ADD COLUMN IF NOT EXISTS "approver_email_bidx" text;--> statement-breakpoint
+ALTER TABLE "pr_merge_log" ADD COLUMN IF NOT EXISTS "triggered_by_email_bidx" text;--> statement-breakpoint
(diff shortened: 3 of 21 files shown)


h3xxit:
Addressed the cubic review in 216f423 — how each finding was resolved:

- **Shape-only ciphertext detection (P1)** — ciphertext now carries an explicit `pii:v1:` version prefix; plaintext can no longer be mistaken for ciphertext.
- **Empty-string ciphertext unparseable (P2)** — `encryptPii('')` stores `''` as itself (no personal data in an empty string, and GCM of an empty plaintext can't be represented in the blob format). This also makes the DB-level `DEFAULT ''` on `change_requests.body` (P3) hold the identical representation as an app-written empty value — default kept, comment added.
- **Concurrent/rolling-upgrade backfill races (both P1s)** — the backfill now runs in one transaction under `pg_advisory_xact_lock` (concurrent boots serialize), and every UPDATE is compare-and-swap: the WHERE pins the values the row was read with, so a writer landing between scan and write makes the update match zero rows — sealed on the next boot pass, never overwritten. UPGRADING.md now says to upgrade with a single app instance (the standard compose flow already does).
- **Boot cost of the every-boot scan (P2)** — the prefix makes "unsealed" a SQL predicate (`NOT LIKE 'pii:v1:%'` / `bidx IS NULL`), so a fully-sealed table costs one empty-result query per boot instead of a full row scan.
- **Normalized-email collisions abort the upgrade (P1)** — handled before the unique index goes up: duplicate approval rows (case-variant email, same logical approval) are collapsed keeping the earliest; duplicate *user* rows are genuine account conflicts and abort the boot (inside the transaction, so nothing half-applies) with the conflicting ids for deliberate operator remediation.
- **Erasure with an undecryptable email (P1)** — fails closed: `eraseUser` refuses before deleting anything when the stored email doesn't decrypt with the configured key, since the email-keyed audit anonymization could not run.
- **Weak domain-separation test (P3)** — replaced with a real assertion: a `TokenCrypto` built from the raw `SECRETS_ENC_KEY` cannot open a PII blob's body.

Re-verified end-to-end against Postgres 18: collision abort + rollback + remediation retry, approval dedupe, prefix on all sealed rows, empty CR body intact, blind-index login/upsert/erasure, idempotent second pass. Full unit suite green (2035 tests).

h3xxit:
Second cubic round, addressed in d1cfdb7:

- **Plaintext beginning with `pii:v1:` evades `needsSealing` (P1)** — fixed: the SQL predicate is now a strict shape regex (prefix + the exact base64 widths GCM produces for the IV and tag), used as a *negative* filter — anything failing it is selected and re-validated app-side by `isEncryptedBlob`, so prefix-impersonating plaintext gets sealed. Sealed rows all match the regex, so steady-state boots still cost one empty-result query per table. Verified E2E: a comment body of `pii:v1: <plaintext>` is sealed and round-trips.
- **Bidx-only CAS can hash a stale/undecryptable source (P2)** — fixed: the update now also pins the source email in its WHERE clause (a concurrent writer replacing the email makes the CAS match zero rows instead of pairing the new email with the old blind index), and fails closed with an actionable error when the source is ciphertext the configured key cannot open, rather than deriving a blind index from ciphertext.
- **Legacy `iv:tag:ct` values need a rewrap path (P1)** — not applicable: that unprefixed format never shipped; it existed only in earlier commits of this same unmerged PR, so no deployed database holds PII ciphertext without the prefix. The secrets vault's unprefixed `iv:tag:ct` values live in columns this backfill never touches.
- **Erasure guard detects by shape, not decryption failure (P2)** — working as intended: a successfully decrypted email is the plaintext address and cannot match the blob shape (base64 has no `@`), while a failed decryption returns the raw blob, which the shape check catches — so shape-of-the-read-value is exactly equivalent to "decryption failed" here.

PR #136, opened 2026-09-04
Git endpoint

aliraza1006:
## Summary by cubic
Adds a remote sync endpoint so a git host or pipeline can tell Hexis the repository changed. Previously a push made outside Hexis showed up only at somebody's next save; now `POST /api/sync/<branch>` pulls that branch's clone and announces the new tree to open browsers, agents, and the skill, tool, and plugin catalogues, all within the request. A branch whose HEAD moved without touching content is reported `updated` but triggers no announcement; a branch deleted on the host reports `remote-gone` and its stale clone is removed.

**New Features**

- `POST /api/sync` accepts Azure DevOps, GitHub/Gitea, and GitLab webhook payloads, an explicit branch list, or no body (syncs every clone).
- Authenticates with `KB_SYNC_SECRET` as a bearer, `X-Gitlab-Token`, or `X-Hub-Signature-256` HMAC, or with an admin's session.
- Concurrent syncs coalesce into one follow-up; each caller's branch list is copied on entry and answered individually, so a hook for branch A never fails on branch B's conflict.
- A pull that landed but couldn't be announced is remembered and delivered by the next sync; the debt is cleared when the branch is found gone or its clone is retired.
- Each branch's outcome (`updated`, `up-to-date`, `not-cloned`, `remote-gone`, `conflict`, `error`) is returned so a pipeline's `curl --fail` fails exactly when Hexis isn't in sync.
- The pull and its observations (HEAD before/after, changed paths) run inside one lock hold, so a concurrent save can't be announced back to its author with the wrong sha.
- Deletion pushes name the deleted branch so its sync reports `remote-gone` and retires the clone; an unborn clone is stale when the branch is gone, a merely empty origin is "nothing to sync", and staged files in an unborn clone are a conflict naming the staged paths.
- Retiring a deleted branch's clone races nothing: `createBranch` now holds the same branch-lifecycle lock as retirement, and retirement backs off while a clone of the branch is bootstrapping, so a branch recreated in that window keeps its clone.
- Conflicts are detected even when `--autostash` exits 0 on a stash conflict, and take the same deterministic rebase-then-background-recovery path; the failed-sync banner links the conflicted files.
- Only finished clones are synced: bootstrapping branches, invalid branch names, and an unreadable workspaces root are never reported as synced.
- Every sync response is stamped `X-Hexis-Sync: result`; the frontend exempts a `503` from the maintenance overlay only when the marker is present, so a reverse proxy's `503` or a `502` still raises it.

**Configuration**

- `KB_SYNC_SECRET` is optional; without it the endpoint only admits an admin's session.
- Setup and Deployment screens show the sync address, the last sync's result, and a "Sync now" button; the panel stays visible even when every knowledge-base setting is env-set.
- New `docs/git-sync.md` covers wiring up GitHub Actions, GitLab/Azure DevOps webhooks, and pipeline steps.

<sup>Written for commit 4bc60093db44158bdb0bf1a3d06466579d74878c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Bevel-Software/Hexis/pull/136?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


h3xxit:
@aliraza1006 — manual review from Razvan's side (head `9cd85c3`). The shape of the feature is right: sync lands as a workflow operation reusing `GitService.pull`, the event bus, and the existing conflict-recovery ladder, with per-branch outcomes instead of throws. The findings are in the details, ordered by severity:

1. **The sync's git operations don't actually serialize with users' git operations** (`workflow.service.ts` ~L534). `listClonedWorkspaces` returns *encoded* directory names (`ali%2Fx`) and the sync hands them straight to `GitService`, while HTTP routes pass the Express-decoded form (`ali/x`) — and `WorkspaceMutex` keys on the raw string. On any slashed branch, a webhook rebase interleaves with the author's concurrent save: index.lock failures, commits landing mid-rebase. `canonicalWorkspaceId()` in `pending-commits.service.ts` exists for exactly this bug class; `syncWorkspaceFromRemote` computes `workspaceIdForBranch(branch)` for its *events* but not for its git calls.
2. **A branch deleted on the host makes sync answer 503 forever** (`sync-payload.ts` ~L104). Deletion pushes aren't skipped (GitHub `deleted:true`, ADO zero `newObjectId`), and a lingering clone of a deleted branch fails its fetch on every subsequent full sync — nothing removes or skips the stale clone. On the documented ADO "PR updated, filter Completed" hook with "delete source branch", every completed PR ends 503, and ADO auto-disables subscriptions after repeated failures. The deleted-branch sweep closes CRs but not clones.
3. **Case-insensitive routing vs case-sensitive raw-body carve-out** (`kb-sync.routes.ts` ~L71). Express matches `POST /api/Sync` to the route, but `isSyncRawBodyPath` says no, so the global JSON parser eats the stream: HMAC callers 401 (no bytes to verify), bearer callers silently sync **all** clones (parsed body isn't a Buffer, branch selection skipped). An odd-cased webhook URL half-works, hiding the misconfiguration. Cheapest: compare case-insensitively; cleaner: stop answering "is this the sync route?" in two places.
4. **`headSha` is read before the pull**, so a clone of an unborn/empty upstream reports `error` + banner where `GitService.pull` deliberately tolerates unborn HEAD (`workflow.service.ts` ~L539).
5. **The read → pull → read → diff sequence takes the mutex four times** (~L599). A user's save landing mid-sequence is broadcast back to them as a "Git sync" file-change with the wrong sha. One `mutex.runAll` section over the whole sequence makes the range exact — and is the natural place to fix (1) too.
6. **One unreadable workspace directory fails every enumerating sync** (`workspace.service.ts` ~L299). The per-entry `.git` probe rethrows non-ENOENT errors (an EPERM/ACL, a half-deleted dir) from inside the loop, so the empty-body webhook form and "Sync now" 500 before any branch is attempted — while your own sync loop (kb-sync.service.ts:113) enforces "one broken clone must never take the other branches down". Push that entry as an `error` outcome instead. (The `/api/sync/<branch>` form doesn't enumerate and is unaffected.)
7. **`CopyValue` re-implements clipboard copy inline** (`SetupScreen.tsx` ~L33) — `lib/clipboard.ts` exists for exactly this and handles the SSR/insecure-context/unfocused cases; the inline copy also loses the `writeText` presence guard.
8. **`setup.api.ts` hand-copies `BranchSyncOutcome`** (~L19) — the union this same PR added to `@bevel-software/platform-shared` so frontend and backend *can't* drift. Import it; core-frontend already depends on the package.

Happy to re-verify any of these if they don't match what you're seeing — and once they're addressed, re-trigger `/bevel-review` with a comment so the architecture check re-runs on the new head.

aliraza1006:
C:/Program Files/Git/bevel-review

h3xxit:
/bevel-review

Re-run under the updated harness and rule set (KB CR #65 + Hexis #138: full-sweep review method, rule-outputs gate, new Git-serialization Invariant). Head unchanged — this run measures the review, not the code.

h3xxit:
/bevel-review

Re-run with the full loop package live: gpt-5.3-codex, the sweep method (KB CR #65), the run-id and rule-outputs gates, and the rule set as of KB CR #69 (Stated Contracts Are Checked; Tests Assert Through Public Interfaces; Git-serialization Invariant). Head unchanged since `aad6c7c6` — this run is the measuring stick.

h3xxit:
@aliraza1006 — manual follow-up review from Razvan's side (head `aad6c7c6`). The response to the first round was the right kind: `resolveWorkspaceDir`, the single-hold sync, the branch-lifecycle lock, and the stale-clone retirement are mechanism-level fixes, and my re-review tried and failed to re-state the original findings against them. The architecture review has already flagged the private-state tests on this head, so these are the remaining findings it and cubic missed:

1. **Coalescing conflates results** (`kb-sync.service.ts` ~L72). A caller coalesced into the follow-up run resolves with the union's result, so `httpStatusFor` answers it for *other callers'* branches: branch A's hook gets a 409 because unrelated branch B conflicts — against the endpoint's own "fails exactly when Hexis is not in sync" contract, and each such 409/503 feeds ADO's auto-disable counter for the wrong subscription. The follow-up run knows which branches each waiter asked for; answer each waiter from its own subset.
2. **`ConflictFiles` interpolates `kbDirName`** (`GitSyncFailedBanner.tsx` ~L23), which the workspace context types and documents as null until bootstrap — a conflict SSE arriving before the workspace-info fetch renders links to `null/<path>`.
3. **The unborn-HEAD arm's premise is false** (`git.service.ts` ~L1622). "No tracked files exist yet" — files *staged* in an unborn clone are tracked, and the `reset --hard` silently destroys them, where the normal path autostashes or surfaces a typed conflict. Either the code honors the claim (stash/refuse when the index is non-empty) or the claim goes.
4. **"Retriable" that can't recover** (`workflow.service.ts` ~L644). A pull that lands but fails to announce returns outcome `error` (503, documented retriable) — but the retry sees `after === before`, reports `up-to-date`, and never re-announces; the missed invalidation is lost for the full TTL while the endpoint says synced. The announcement has to be owed until delivered (e.g. record the pending announce with the pull, emit on the next observation), not implied by sha movement.
5. **`secretEquals` re-implements `timingSafeStringEqual`** (`sync-auth.ts` ~L64) — already exported from `modules/auth/password-hash.ts` with the same length-hiding property. Import it; two constant-time compares means a future timing-safety audit fixes one and misses the other.

Findings 1, 3 and 4 are the same review move — the diff's own stated contracts tested against the code — so worth one design pass together rather than three patches (see the `address-review-bot-findings` skill in the KB).

Also: your `/bevel-review` trigger comment got mangled by Git Bash into `C:/Program Files/Git/bevel-review` (MSYS path conversion), so that run never started — prefix with `MSYS_NO_PATHCONV=1` or comment from the GitHub UI, and feel free to delete the mangled comment. The current errored review status on this head is a harness false positive, fixed in #142; it'll clear on the next re-trigger after that merges.

h3xxit:
/bevel-review

Re-trigger after the 15:41 run died silently mid-review — the knowledge base was briefly unreachable at that time (confirmed independently); it is healthy again. Rule set now includes KB CR #70 (mechanical claim sweep + citations on every verdict).

h3xxit:
/bevel-review

Re-trigger on pipeline-agent 0.1.7 — a severed model connection mid-turn now recovers with a queued retry (or fails loudly) instead of ending the run silently. Rule set unchanged since the last attempt (through KB CR #70).

h3xxit:
/bevel-review

Re-run: the 08:00 bounce was an evidence race — cubic was still pending on `4bc60093` when the review checked; it has since completed green, as has `test`.

PR #161, opened 2026-09-09
Grep-File-Path — Grep given a file path searches that file, never a quiet empty

h3xxit:
Fixes ticket: `Data/Engineering/Knowledge/Hexis/Tickets/Grep-File-Path.md`

## Summary
`grep` now treats a path naming a readable file the same way `read_file` and the directory walk treat it, instead of silently returning an empty match set:

- A path naming a readable text file is searched directly: matches carry that path, 1-based line numbers, and the same match shape/cap behavior as a directory grep.
- A path naming a readable document file (office/PDF/email) is searched via the same extraction the directory walk uses — marker lines included, extraction budget and the existing `note` for unextracted documents respected.
- A path that exists but has no searchable text (image, binary with NUL bytes, corrupt document) returns empty matches plus a `note` explaining there is no searchable text.
- A path that does not exist fails with a structured error naming the path, never an empty success.
- A file path the caller may not read responds exactly as `read_file` does for that same path (same error and shape), so `grep` and `read_file` tell one story.
- Directory grep behavior is unchanged: unreadable entries are still silently filtered from walks, and depth/match caps are unchanged.

## Testing
- `packages/core-backend` vitest: 195 files / 2340 tests passed.
- `tsc --noEmit` clean for all packages except `apps/server` (pre-existing unbuilt-dist error, unrelated to this change).
- Live verification against a from-source build (`deployment/docker-compose.build.yml`) covering all acceptance criteria: file hit, file miss, cap/truncation, document (pptx) extraction with marker lines, document extraction budget note, no-searchable-text note (binary/image/corrupt doc), missing path error, unreadable file parity with `read_file`, and an unchanged directory-grep regression.

---
## Summary by cubic
Fixes the Grep-File-Path ticket: `grep` given a path naming a readable file now searches that file directly, instead of silently returning an empty match set.

- Missing paths fail with a structured 404 naming the path, including paths under an existing file (ENOTDIR).
- Files with no searchable text (image, binary, corrupt document) return empty matches plus a `note`.
- Any stat failure that isn't absence routes as a file, so the read gate answers first: an unreadable path gets `read_file`'s exact 403, and a readable but unresolvable path fails with `read_file`'s own error — nothing is invented or disclosed.
- Document files go through the same extraction as the directory walk, marker lines and extraction budget note included.
- Directory grep is unchanged: unreadable entries still silently filtered, depth and match caps unchanged.

<sup>Written for commit b338e2f4a0e6f80b91ed23c1548d7d30d248eb88. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Bevel-Software/Hexis/pull/161?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/core-backend/src/modules/workspace/__tests__/workspace.tools.test.ts
+++ b/packages/core-backend/src/modules/workspace/__tests__/workspace.tools.test.ts
@@ -599,6 +599,22 @@ describe('grep with a path that names a file', () => {
     expect(await grepRes.json()).toEqual(await readRes.json());
   });
 
+  it('a READABLE path the filesystem cannot resolve fails exactly as read_file fails', async () => {
+    const loop = `${KB_DIR}/Knowledge/Tangle.md`;
+    const base = await start();
+    await mkdir(join(tempDir, KB_DIR, 'Knowledge'), { recursive: true });
+    await symlink('Tangle.md', join(tempDir, loop));
+    const [readRes, grepRes] = await Promise.all([
+      post(`${base}/api/agent/tools/read_file`, { path: loop }),
+      post(`${base}/api/agent/tools/grep`, { pattern: 'needle', path: loop }),
+    ]);
+    // The gate allows it, so the READ is what answers — and it is the same
+    // `fs.readFile` read_file calls. grep must not dress that up as absence
+    // (a false 404), nor invent a failure of its own: one path, one story.
+    expect(grepRes.status).toBe(readRes.status);
+    expect(await grepRes.json()).toEqual(await readRes.json());
+  });
+
   it('a DIRECTORY path still walks the whole subtree (unchanged)', async () => {
     const base = await start();
     await fs.writeFile('notes/one.md', 'needle here\n');
--- a/packages/core-backend/src/modules/workspace/__tests__/workspace.tools.test.ts
+++ b/packages/core-backend/src/modules/workspace/__tests__/workspace.tools.test.ts
@@ -1,6 +1,6 @@
 import type { Server as HttpServer } from 'node:http';
 import { execFile } from 'node:child_process';
-import { mkdir, mkdtemp, readFile, rm } from 'node:fs/promises';
+import { mkdir, mkdtemp, readFile, rm, symlink } from 'node:fs/promises';
 import { tmpdir } from 'node:os';
 import { join } from 'node:path';
 import { promisify } from 'node:util';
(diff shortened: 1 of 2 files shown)


h3xxit:
Both cubic findings were valid and are fixed in `65c7483f`.

**P2 — `ENOTDIR` is absence too** (`workspace.tools.ts:292`). Agreed. A path whose parent is an existing file (`notes/deep.md/deeper.md`) fails its stat with `ENOTDIR`, and nothing can ever live there, so it is exactly as absent as `ENOENT`. `searchRootKind` now treats both codes as `missing`, and the path gets the documented 404 naming it instead of a raw failure.

**P3 — a stat failure must not outrank the read gate** (`workspace.tools.ts:656`). Agreed on the substance. `searchRootKind` no longer throws; it returns `{ kind: 'unknown', err }` and lets the handler decide. The handler runs `assertCanRead` first, so a denied path answers 403 identically to `read_file` even when the filesystem could not say what is there (EACCES, ELOOP, I/O). Only after the gate lets a path through is the underlying error surfaced — which is the same story `read_file` tells, since its own `readFile` would raise it.

One deliberate deviation from the suggested "resolve the kind after the gate": the stat still comes first. grep cannot apply a *file* gate until it knows the root is not a directory — a caller-denied DIRECTORY must keep returning an empty walk rather than a 403, which is the unchanged-directory-behavior acceptance criterion. So the ordering is: stat (no verdict drawn from it) → gate → existence/extraction. The stat leaks nothing, because what the caller is told is the gate's verdict, never the filesystem's.

Two regression tests added, both verified to fail against the previous head:
- a path under an existing file returns 404 naming it (`ENOTDIR`);
- a denied path that is a self-referential symlink (stat fails `ELOOP`) returns a byte-identical 403 from `grep` and `read_file`.

`workspace.tools.test.ts`: 80/80 passing; `tsc --noEmit` clean for core-backend.

h3xxit:
Re the new P2 on `throw root.err` (`workspace.tools.ts:698`) — the concern is fair, but the suggested fix would have made things worse, so I took a different route. Addressed in `b338e2f4`.

**Why not a sanitized `ToolError` here.** I measured it. For a readable path the filesystem cannot resolve (a self-referential symlink), the two tools returned *byte-identical* bodies:

```
read_file  500 {"error":"ELOOP: too many symbolic links encountered, realpath '/tmp/ws-tools-7yY65O/knowledge-base/Knowledge/Loop2.md'"}
grep       500 {"error":"ELOOP: too many symbolic links encountered, realpath '/tmp/ws-tools-7yY65O/knowledge-base/Knowledge/Loop2.md'"}
```

So wrapping only grep's throw would have (a) broken this PR's central contract — grep and `read_file` telling one story for the same path — and (b) closed nothing, since one `read_file` call on that same path hands the caller the identical absolute path. The disclosure is real, but it is a property of the shared handler boundary (`tool-handler.ts` returns `err.message` verbatim as a 500 for any non-`ToolError`), not of this code.

**What I did instead — delete the special case.** The `unknown` outcome only ever existed to carry a stat failure back for the handler to rethrow, and that rethrow was the sole place grep raised raw OS text from a call `read_file` never makes. `searchRootKind` now returns plain `'directory' | 'file' | 'missing'`: a stat failure that is not absence simply answers `'file'`, sending the path down the ordinary single-file route. The read gate still speaks first (a denied path is 403, unchanged), and a readable one is then answered by the *same* `fs.readFile` that `read_file` calls. grep now tells `read_file`'s story by **construction** rather than by coincidence, and the flagged `throw root.err` is gone — along with one branch, one union member and one rethrow. Net −24 lines in the handler.

New test pins it: a readable unresolvable path must produce a byte-identical status and body from `grep` and `read_file` — grep may neither dress the failure up as a false 404 nor invent one of its own.

**Flagged for its own ticket, not this one:** sanitizing raw filesystem text at the `tool-handler.ts` 500 boundary. It is a genuine hardening, but it is cross-cutting (every workspace tool) and a real product trade-off — today an agent that gets `File not found: ok.md` learns something actionable, and a blanket "an internal error occurred" would cost that. Happy to do it here instead if you would rather it land in this PR.

core-backend: 81/81 in `workspace.tools.test.ts`, `tsc --noEmit` clean.

h3xxit:
C:/Program Files/Git/bevel-review

PR #167, opened 2026-09-10
MCP-Session-Recovery — MCP tool calls survive a server restart

h3xxit:
Ticket: `Data/Engineering/Knowledge/Hexis/Tickets/MCP-Session-Recovery.md`

## Summary

When an MCP server (the platform's own `hexis-mcp` manual, or a proxied third-party server declared in `mcp.json`) restarts and drops its session, the next tool call now recovers transparently instead of failing: the caller sees success (or the tool's own genuine error) and nothing else, with a single log line noting the recovery.

- Added `packages/mcp-core/src/session-recovery.ts`: a shared recovery mechanism used by both entry points — tool calls dispatched over the MCP surface and tool calls made from inside a code-mode chain (`call_tool_chain`). One mechanism, not two.
- Recovery triggers **only** on session-not-found-shaped failures (HTTP 404 + JSON-RPC `-32001` / "Session not found" on a request that carried a session id). Any other failure (genuine tool error, auth failure, timeout, other transport error) is surfaced unchanged and is never retried — pinned by a test.
- Exactly one re-registration and one retry per failed call; if the retry also fails, that failure reaches the caller.
- Concurrent calls that hit session loss on the same manual coalesce into a single re-registration (no re-register stampede).

## Changed packages

`@bevel-software/platform-mcp-core`, `@bevel-software/hexis-mcp`, `@bevel-software/platform-core-backend`

## Tests

- `packages/mcp-core/src/__tests__/session-recovery.test.ts` (new) + `fake-mcp-server.ts` test double
- mcp-core 101/101, hexis-mcp 136/136, core-backend 2357/2357 passing
- typecheck + build clean for all three changed packages
- Covers: platform-restart recovery, third-party-server recovery, non-session errors not retried, single-flight coalescing, chain-path (`call_tool_chain`) recovery, retry-failure surfacing

A changeset is included (`.changeset/mcp-session-recovery.md`).

Unrelated pre-existing failure noted for the record: `packages/core-frontend` vitest fails entirely with `React.act is not a function` (1266 tests) — reproduced with this branch's changes stashed, so it pre-exists and no frontend file was touched here.

---
## Summary by cubic
Makes MCP tool calls survive a server restart: when a server drops its session, the next call re-registers the affected manual and retries once instead of failing with the spec's 404 / "Session not found". Previously a redeploy left connected agents' tools erroring until someone reconnected by hand.

**Recovery behavior**
- Only session loss is retried — HTTP 404 with JSON-RPC `-32001` / "Session not found", which a server decides before dispatching to a tool.
- Tool errors, auth failures, and timeouts pass through unchanged and are never retried, so a mutating tool never executes twice.
- Concurrent session losses on one manual share a single re-registration; a failed retry or re-registration surfaces the caller's original error.
- Recovery re-checks the session generation right before deregistering, so a session replaced while it waited is reused, not torn down.
- In `hexis-mcp`, recovery and credential renewal share one re-registration gate; shutdown awaits it before teardown, and calls parked behind it get ConnectionClosed.
- Applies to the hosted proxy, local `hexis-mcp`, and proxied third-party servers, on both the MCP dispatch surface and `call_tool_chain`.

<sup>Written for commit 61d1a5d9fd1124d5771e0510abb1fc22e35778a4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Bevel-Software/Hexis/pull/167?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/mcp-session-recovery.md
+++ b/.changeset/mcp-session-recovery.md
@@ -1,0 +1,7 @@
+---
+'@bevel-software/platform-mcp-core': patch
+'@bevel-software/platform-core-backend': patch
+'@bevel-software/hexis-mcp': patch
+---
+
+MCP tool calls survive a server restart. A restarted server answers the next call on an old session with the spec's 404 / `Session not found`, and our clients used to hand that straight to the caller — a platform redeploy left every connected agent erroring until someone reconnected it by hand. The client now re-registers the affected manual and retries the call once, so the restart shows up as a single log line instead of a broken toolset. This covers third-party MCP servers the workspace proxies as well as the platform itself, and the local `hexis-mcp` server as well as the hosted endpoint. Only that one failure — the session miss, which the server decides before it dispatches to a tool — is retried: a tool error, an auth refusal, a timeout or a dropped connection surfaces exactly as before, so nothing that could have already run is ever run twice.
--- a/packages/core-backend/src/modules/mcp/mcp.service.ts
+++ b/packages/core-backend/src/modules/mcp/mcp.service.ts
@@ -28,6 +28,7 @@ import {
   dispatchMetaTool,
   dispatchToolCall,
   registerManual,
+  installSessionRecovery,
   flattenManualTool,
   toListedTool,
   toolError,
@@ -447,7 +448,24 @@ export class McpService {
       variables,
       load_variables_from: [bevelSecretsLoaderConfig(userId)],
     });
-    return CodeModeUtcpClient.create(process.cwd(), config);
+    const client = await CodeModeUtcpClient.create(process.cwd(), config);
+    /**
+     * SESSION RECOVERY. A proxied third-party MCP server that restarts (or
+     * expires a session) answers our next call with the spec's 404/`-32001`;
+     * without this the session's tools stay broken until the caller reconnects.
+     * Installed on the client, so the dispatch surface and `call_tool_chain`
+     * recover through the same mechanism. Re-registration re-uses the very
+     * template discovery registered with, resolved through the same variable
+     * tiers, so a healed manual is identical to a freshly discovered one.
+     *
+     * Deliberately clear of {@link ManualFailureMemo}: that memo is a circuit
+     * breaker for manuals whose REGISTRATION failed — which have no discovered
+     * tools to call in the first place — so a recovery neither consults it (a
+     * memoized failure must not block healing a manual that registered fine)
+     * nor records into it (an evicted session is not a broken credential).
+     */
+    const byName = new Map(manuals.map((m) => [String(m.name), m]));
+    return installSessionRecovery(client, { manualTemplate: (name) => byName.get(name) });
   }
 
   /**
--- a/packages/hexis-mcp/src/server.ts
+++ b/packages/hexis-mcp/src/server.ts
@@ -323,26 +323,61 @@ export async function createHexisMcpServer(
    * template from the repository BY NAME at call time, so re-registration is
    * invisible to it (verified against @utcp/sdk's dispatch).
    *
+   * SESSION RECOVERY SHARES THIS GATE. Recovery (below) re-registers a manual
+   * too, so both go through `withReregisterGate`: one at a time, never
+   * interleaved, and never each waiting on the other. That last part is why
+   * `callsParkedForReregister` exists — a call queued at the gate still holds
+   * an `inflightCalls` slot, and a swap draining for it would wait out the
+   * full deadline for a call that is itself waiting for that swap.
+   *
    * Key mode sets no `renewConnectionKey`, so renewal.ts never renews, this
-   * listener is never called, and the gate below never engages.
+   * listener is never called, and the swap below never engages.
    */
   let closed = false;
   let remoteManualRegistered = false;
   let inflightCalls = 0;
-  let swapInProgress: Promise<void> | null = null;
+  /** The one re-registration allowed at a time: a credential swap or a recovery. */
+  let reregisterInProgress: Promise<void> | null = null;
+  /** In-flight calls parked at that gate, waiting for their turn to recover. */
+  let callsParkedForReregister = 0;
   const sleep = (ms: number): Promise<void> => new Promise((resolve) => setTimeout(resolve, ms));
+  /**
+   * Run `body` as THE re-registration in flight. `parksACall` marks the waiter
+   * as a tool call whose own dispatch is blocked here, so a swap's drain does
+   * not count it. The published promise never rejects: it says "finished", not
+   * "succeeded".
+   */
+  const withReregisterGate = async <T>(body: () => Promise<T>, parksACall = false): Promise<T> => {
+    if (parksACall) callsParkedForReregister += 1;
+    try {
+      while (reregisterInProgress) await reregisterInProgress;
+    } finally {
+      if (parksACall) callsParkedForReregister -= 1;
+    }
+    // `body()` runs to its first await before the publish below, and nothing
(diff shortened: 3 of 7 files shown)


h3xxit:
All four findings from the last cubic review are addressed in code, in commit `4570456c`. Summary, in review order:

**1. `session-recovery.ts:167` — a rejecting `manualTemplate` replaced the caller's error. Accepted, fixed.**
The re-registration is now split in two: `reregisterNow` does the work, and `reregister` wraps *the whole thing* — template resolution, deregister/register, cleanup, and the new surface gate — in one `try/catch` that turns any throw into `{ ok: false }`. Recovery therefore never manufactures an error of its own: the caller always sees the original session loss. Pinned by `surfaces the original session loss when the template resolver throws` (asserts the caller's error is the `-32001` one, and that the call was *not* retried) and by `surfaces the original session loss when the surface gate throws`.

**2. `session-recovery.ts:229` — one recovered call could emit several log lines. Accepted, fixed.**
`reregisterNow` no longer logs; it collects `notes` (deregister failed, re-registration failed, post-registration cleanup failed) and `recover` folds them into the single line the call is allowed to emit — `… — re-registered and retried. (deregistering first failed: …; post-re-registration cleanup failed: …)`, or `… — not recovered (…); the original failure stands.` when it could not recover. A manual this surface holds no template for stays silent, as before: nothing happened, so there is nothing to report. Pinned by `keeps one recovered call to one log line however much went sideways`.

**3. `server.ts:406` — a credential swap and a recovery could wait on each other for the 15s drain. Accepted, fixed — both remedies you suggested, since each covers a different half.**

- *One single-flight operation*: `mcp-core` grew an optional `withReregister(manualName, run)` hook, and `hexis-mcp` passes the same gate the swap uses (`withReregisterGate`). A swap and a recovery are now the same serialized slot, so they can no longer interleave — which also closes a window the old code had, where a recovery's deregister/register pair ran outside the dispatch gate and a concurrent call could look up a manual that was momentarily absent. `manualTemplate` no longer waits on the swap itself; under the gate that would be self-deadlock.
- *Exclude the recovering call from the drain*: a call parked at the gate still holds an `inflightCalls` slot but is, by definition, waiting for the very operation that is draining. `callsParkedForReregister` counts those, and the drain condition is now `inflightCalls - callsParkedForReregister > 0`. Without this, serializing alone would have preserved the stall in the other order (swap first, recovery arriving second).

One consequence worth stating plainly: if a swap re-registers while a recovery is parked, the recovery still re-registers when its turn comes. That is one extra dial, fully serialized, and it is deliberate — it is what makes the retry provably run against a session created after the failure rather than one the recovery would have to reason about.

**4. `server.ts:407` — shutdown could race a recovery past the `closed` check. Accepted, fixed.**
The gate is published before the re-registration body runs, and `shutdown` awaits it (`if (reregisterInProgress) await reregisterInProgress`) after setting `closed`. Both orderings are now closed: a recovery that already holds the gate finishes before `client.close()` runs — including any local manual it registered, whose spawned child that close then tears down — and a recovery still queued behind the gate finds `closed` in `manualTemplate` and registers nothing at all. The gate promise never rejects, so this cannot throw teardown off course.

Tests: `platform-mcp-core` 105/105 (`session-recovery.test.ts` 16 → 20), `hexis-mcp` 136/136, `core-backend` mcp module 122/122; typecheck clean across all three, and eslint reports nothing on the changed files. The gate contract itself is pinned in `mcp-core` (`runs the whole re-registration inside the surface gate, once`) rather than re-tested end to end in `hexis-mcp`, whose stub deployment is stateless streamable-HTTP and so has no session to lose.

h3xxit:
Both findings from the second cubic review are addressed in code, in commit `4cd010d2`.

**1. `server.ts:353` — a call parked at the gate dispatches after shutdown closed the client. Accepted, fixed.**
The check went into the `CallTool` handler rather than into `withReregisterGate`, which is where the symptom is and what your comment text describes ("before incrementing and dispatching"): the gate is also the swap's and the recovery's path, and neither wants a throw there. After the wait the handler now does `if (closed) throw new McpError(ErrorCode.ConnectionClosed, 'hexis-mcp is shutting down.')`, so a call that loses the race gets the actual reason instead of whatever a half-torn-down transport happens to raise. `shutdown` deliberately still does not wait for callers parked behind the gate — refusing them is the cheaper and more honest answer than draining them.

**2. `server.ts:438` — recovery re-registers a manual the renewal already replaced. Accepted — you were right and my previous note calling it deliberate was wrong.**
Coordinated through the generation counter, as you suggested:

- `mcp-core` exports `noteManualReregistered(client, manualName)`, which bumps that manual's generation. `hexis-mcp`'s credential swap calls it after a successful re-registration, so a re-registration that did NOT come from recovery is still visible to recovery.
- The generation is now re-checked *inside* the gate. It was already checked before the attempt, but that is too early for this race: by the time a recovery reaches the gate, the swap holding it has not finished yet. `reregisterNow` takes the generation the failing call saw and returns `{ ok: true, reused: true }` when it has moved, so the call retries against the swap's session and never deregisters it.

The reverse ordering (recovery first, swap queued behind) still re-registers: the swap is the credential authority, and a token that arrived mid-recovery has to become effective somewhere. That is one serialized dial with no session thrown away, not the wasteful case you flagged.

Pinned by `reuses a session the surface re-registered while the recovery waited at its gate` — it asserts zero `registerManual` and zero `deregisterManual` calls, one retry, and the single `— re-registered by a concurrent call; retrying.` line — plus a no-op test for a client that never installed recovery.

Tests: `platform-mcp-core` 107/107 (`session-recovery.test.ts` 20 → 22), `hexis-mcp` 136/136, `core-backend` mcp module 122/122, typecheck clean for all three, eslint clean on the changed files. The shutdown refusal is not covered by a new in-process test — driving it needs a stub deployment that both serves a tool and can hold the gate open, and the behaviour is a two-line guard on a flag that `shutdown` already sets; the runtime pass on the previous head exercised the shutdown-during-recovery path directly (client.close() at 1/5/20/60ms into a recovery: returned in 32–52ms, no orphaned stdio children).

h3xxit:
The finding on `4cd010d2` is addressed in code, in commit `61d1a5d9`.

**`session-recovery.ts:231` — the generation is read before the `manualTemplate` await, so a swap landing during it is missed. Accepted, fixed — and a fair catch: my previous fix moved the window rather than closing it.**

`manualTemplate` is documented as a hook a surface may park in (waiting out a re-registration of its own is one of the reasons it is allowed to be async), so the counter can move across that await exactly as it can across the gate. `reregisterNow` now tests it in both places, and the comment says which one is load-bearing: the check immediately before the deregister decides, the earlier one only skips work nobody needs.

Pinned by `reuses a session the surface re-registered while the template resolver waited` — the resolver parks on a deferred, the test reports a surface re-registration while it is parked, then releases it, and asserts zero `registerManual` and zero `deregisterManual` calls with the retry still succeeding. I checked it fails against the previous head's code (`registrations` is 1 instead of 0), so it pins the fix rather than passing either way.

Tests: `platform-mcp-core` 108/108 (`session-recovery.test.ts` 22 → 23), `hexis-mcp` 136/136, `core-backend` mcp module 122/122, typecheck clean for all three, eslint clean on the changed files.

PR #180, opened 2026-09-14
Agents never create app roles — a "new role" ask is usually a group

h3xxit:
Ticket: `Data/Engineering/Knowledge/Hexis/Tickets/Agents-Never-Create-Roles.md`

## Summary

App roles are pre-set; agents should never be able to invent a new one through a file-tool write to `roles.yaml`. This PR makes any agent-driven write that would introduce a role name not already present in the file fail closed, while leaving membership edits and all human paths untouched.

- Agent `roles.yaml` writes (`write_file`, `write_files`, `edit_file`, `append`, and `copy_file`/`move_file` landing on `roles.yaml`) now refuse any candidate containing a role name absent from the current file, returning a 422 `RolesYamlNewRoleError` (`kind: roles-yaml-new-role`, with the offending `roleNames`) and pointing the caller at using a GROUP for a task- or team-scoped set of people. Nothing is written on refusal, and a multi-file batch fails closed as a whole.
- Writes that only add/remove/reorder members of *existing* roles continue to succeed exactly as before.
- A rename reads as delete+create and is refused for the created name — no special-cased rename logic.
- `unzip` never extracts a `roles.yaml` entry.
- The editor's `PUT /file` save path and the App-roles admin service are unaffected — they keep the existing parse-only validator and can still create/delete/rename roles.
- The `AGENTS.md` platform-owned template gains a roles-vs-groups decision guide (what roles are, why agents never create them, an "is it really a group?" test, and what to do instead), propagated to deployments the same way other template updates propagate.

## Testing

- `pnpm install`; `core-backend` `tsc --noEmit` clean; `vitest run` — 209 files / 2647 tests passed.
- Booted the full stack from source via `deployment/docker-compose.build.yml` and exercised every acceptance criterion end-to-end (new-role write refused with the group-redirect message and file left unchanged; members-only writes land; editor save and App-roles service still create/rename/delete roles; rename refused for the new name while plain delete succeeds; same single-error 422 shape as the existing invalid-`roles.yaml` refusal; template boot delivers the updated `AGENTS.md` guide).

---
## Summary by cubic
Prevents agents from creating app roles: any agent file-tool write to `roles.yaml` that adds a role name not already in the file is refused with a 422 `RolesYamlNewRoleError` that redirects to using a group. Membership-only edits land as before, and human paths (the editor's `PUT /file` save and the App-roles admin service) keep the existing parse-only validator.

**Behavior**
- The gate covers `write_file`, `write_files`, `edit_file`, `append`, `copy_file`, and `move_file` onto `roles.yaml`; a multi-file batch fails as a whole and nothing is written.
- A rename reads as delete plus create, so it is refused for the created name; `unzip` never extracts a `roles.yaml` entry.
- Writes are judged before the lock and again under it against the bytes that land, so a concurrent change while waiting can't reinstate a deleted role or falsely refuse a valid one; any refusal under the lock releases untouched, no matter what was thrown, and a move releases both its locks.
- `AGENTS.md` gains a roles-vs-groups guide that is propagated to deployments on boot.

<sup>Written for commit 6d25ea4bcc93c5cff82d88a555271c0fd075bae4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Bevel-Software/Hexis/pull/180?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


h3xxit:
Addressed all three cubic findings in b3c6f18e.

**P1 — `locking-filesystem.ts:161` (validated pre-lock bytes ≠ written bytes)** — accepted. The content gate still runs before the lock (an ordinary refusal takes no lock), and now runs again once the lock is held; that second check is the one that counts:
- `writeFile` / `writeFiles` re-judge the exact bytes they write, after every caller lock is acquired.
- `appendFile` judges `current + appended` under the lock and, for a claimed path, writes exactly those bytes.
- `copyFile` (source is not locked) reads the source under the destination lock, judges it, and writes exactly those bytes.
- `moveFile` is re-judged with both of its locks held, so the source cannot change before the move lands.
- A refusal under the lock releases **untouched** (`releaseLockUntouched`), for every enclosing lock too — never `releaseLockNoCommit`, whose reset to HEAD would destroy a prior save's still-queued bytes.

**P2 — `roles-yaml-guard.ts:166` (current roles.yaml read outside the lock)** — accepted, and fixed by the same change: the agent validator's `readCurrent` now also runs under the roles.yaml lock, so a role deleted while the write waited cannot be reinstated, and a role added meanwhile is not falsely refused by the authoritative check. (The early pre-lock check could in theory still refuse a role added in the milliseconds before the lock, but only for a candidate that names that exact new role, i.e. a write that was trying to create it; a retry passes.)

**P3 — `steps.test.ts:211` (test coupled to template line breaks)** — accepted. The section is now found by its own heading up to the next heading of any level (or the end of the file), and the assertions run on whitespace-collapsed prose, so re-wrapping the text or reordering headings doesn't break the test.

New tests (`locking-filesystem.test.ts`, *judged again under the lock*) make the lock acquire simulate a concurrent writer for writeFile, writeFiles, appendFile, copyFile and moveFile, and assert: refused, every lock released untouched, nothing committed, roles.yaml unchanged; plus a members-only write that still lands.

Verification: core-backend `tsc --noEmit` clean; vitest over kb-fs, access-model, access, tool-helpers, workspace — 65 files / 1104 tests passed.

h3xxit:
Second cubic pass: both open findings are addressed in 6d25ea4b.

**P2, `locking-filesystem.ts:763` (a check that throws a primitive fell through to the no-commit release):** accepted. The old code recognised a refusal by adding the thrown *object* to a WeakSet, so a string or other primitive never got marked, and the lock was released via `releaseLockNoCommit`, which resets the file to HEAD. Now:
- A `check` failure leaves the raw lock cycle (`lockedCycle`) wrapped in a `CheckRefusal`, whatever value was thrown, and that cycle releases untouched.
- `moveFile`'s inner lock uses the raw cycle, so the outer lock also sees the refusal and both release untouched.
- `withLock` unwraps the refusal, so the caller still gets exactly the value the check threw.
- New tests throw a bare string from the under-lock check for `writeFile` (one lock) and `moveFile` (two nested locks). They assert the caller receives that string, every lock is released untouched, and `releaseLockNoCommit` is never called.

**P2, `roles-yaml-guard.ts:166` (current roles.yaml read outside the lock):** this was already fixed in behaviour by b3c6f18e. The validator's `readCurrent` is not cached and runs on every call, and `LockingFilesystem` now calls the validator a second time with the `roles.yaml` lock held, for write, write_files, append, copy and move. The race tests in `locking-filesystem.test.ts` (*judged again under the lock…*) use this real validator and cover both directions: a role deleted while the write waited is not reinstated, and a members-only write still lands after a concurrent change. The finding stayed pinned only because that line didn't change, so 6d25ea4b adds a doc comment to `makeAgentRolesYamlWriteValidator` stating the contract: fresh read on every call, and the call under the lock is the one that decides.

Verification: core-backend `tsc --noEmit` clean; vitest over kb-fs, access-model, access, tool-helpers and workspace: 65 files / 1106 tests passed.

PR #197, opened 2026-09-15
Repo-Folders-Checked-At-Setup — Setup shows the three root folders up front and checks them against the repository

h3xxit:
Ticket: `Data/Engineering/Knowledge/Hexis/Tickets/Repo-Folders-Checked-At-Setup.md`

## Summary

Setup shows the Knowledge, Skills and Plugins folder fields in the main "Knowledge, skills & tools" section instead of under Advanced, in both the first-run and Deployment variants, with their existing help text and validation (three distinct names).

Test connection now lists the repository's top-level folder names on the default branch (blob-less partial clone with a plain shallow clone fallback, cleaned up on every path) and offers them as suggestions on the three folder fields, the same way branch names are suggested. Each configured root reports one of: found; not in the repository yet — it will be created; or a case/trailing-`s` variant warning naming the actual folder and how to resolve it. An empty repository shows the single existing "empty; it will be set up for you" message instead of three will-be-created warnings. Warnings never block Save.

On the save that completes first-run setup, the three folder names are applied to the running process before the initialization phase runs (fixing a bug where `TemplateFilesStep` and `buildSeedTree` had snapshotted the default root names at construction, so the phase scaffolded `KnowledgeBase/`/`Skills/` alongside the customer's configured folders). The phase now resolves the core roots per run, so the app serves the configured names without a restart. Later changes on the Deployment page stay restart-to-apply as today.

## Tests

- Per-root states (found / will-be-created / case-variant, including trailing-`s`), empty-repository message, suggestions, non-blocking Save, field placement in both variants.
- `setup.routes.layout-phase.test.ts`: end-to-end completing-save test through a real `KbStartupRunner` + `TemplateFilesStep`, asserting the pushed tree holds the configured names and not the defaults.
- Listing command shape and temp-dir cleanup on every path.

---
## Summary by cubic
Setup now puts the Knowledge, Skills, and Plugins folder fields in the main section and checks them against the repository during Test connection. Completing first-run setup applies the selected names before initialization, so repositories using non-default roots no longer get duplicate default folders or need a restart.

**Repository checks**
- Test connection lists top-level folders from the remote’s default branch using a blob-less clone with a shallow-clone fallback; empty repositories return no folders, and listing failures do not fail the connection.
- Fields report found, will be created, or case/trailing-`s` variants, offer valid folder names as suggestions, and never block Save.

**Setup initialization**
- The completing save applies the folder names and branch model to the running process; retries after failed initialization use corrected names, while later Deployment changes remain restart-to-apply.
- Startup seeding and template steps resolve the active folder layout when they run instead of when they are created.

<sup>Written for commit dfb39645223b234aaabd6ef992e7fc31f8954dbe. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Bevel-Software/Hexis/pull/197?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


h3xxit:
Responses to the cubic review of `9986aead`. Fixes are in `2a98de60`, which is committed and will be pushed with the next pass.

1. **`kb-layout.ts:193`, a failed first-run retry ignored corrected folder names.** Accepted. The route now records that setup applied the layout. While that run still stands failed, a retrying save re-applies the stored names. The app gate never opened, so nothing live is disturbed. New test: *applies names corrected between a failed run and its retry*.
2. **`git-root-folders.test.ts:184`, `file://` URL on Windows.** Accepted: the URL is now built with `pathToFileURL(origin).href`.
3. **`root-folders.test.ts:24`, the reverse plural case was not fully asserted.** Accepted: every variant case now asserts the full state with `toEqual`.
4. **`git-root-folders.ts:142`, stat before removing `target`.** Declined. `target` is always `<mkdtemp dir>/repo`: a fresh private directory this call created, with a fixed child name. Anything at that path was put there by the refused `git clone` a moment earlier. `rm({recursive, force})` removes a file or an empty path equally well and does not follow symlinks. Refusing a non-directory would turn a harmless leftover into a failed listing, without protecting any path outside the call's own scratch directory. The clone-path protection the team's feedback refers to guards caller-influenced paths, and this one is not.
5. **`git-root-folders.ts:154`, raw git error text in the log.** Accepted: the line now goes through `printable()` after the token is redacted. There is a test with C0 and C1 (CSI) bytes.
6. **`root-folders.ts:24`, a plural-only difference was reported as a case difference.** Accepted: the state now carries `difference: 'case' | 'trailing-s' | 'case-and-trailing-s'`. The messages read "differs only by case", "differs only by a trailing s" and "differs by case and a trailing s". There are unit and UI tests for each.
7. **`SetupScreen.tsx:686`, invalid suggestions.** Accepted: listed folders are filtered through `validateKbRootName` and the reserved roots (`Data`, `Agents`, `Pipelines`, compared case-insensitively) before they are offered. The state check still sees the full listing.
8. **`setup.routes.ts:367`, `ls-remote --heads` never reports `HEAD`.** Accepted and confirmed against git 2.39. The check now runs `ls-remote --symref --end-of-options <url> HEAD refs/heads/*`, parsed by `parseLsRemote`. A test against a real repository whose trunk is `production` checks that the listing branch is `production`.
9. **`deployment-settings.service.ts:474`, `restartRequired` over a branch model applied live.** Accepted: when the save applies the branch model to the running process, `defaultBranch` and `protectedBranches` no longer count towards `restartRequired`, the same as the folder names. There are tests for both the applied case and the already-configured case, which still asks for a restart.

Backend settings and startup tests: 147/147. Frontend setup tests: 69/69. `tsc` is clean in core-backend and core-frontend, and eslint is clean on the changed files.

h3xxit:
Follow-up on the three findings cubic's summary still lists as open. Cubic skipped `c6459e29` because it's a merge commit, so none of these have been re-reviewed since the fixes landed.

- **`kb-layout.ts:193`, retry after a failed first run.** Fixed in `2a98de60` and carried through the merge. The fix lives in `setup.routes.ts`, not on the anchored line: the route records `layoutAppliedBySetup` and re-applies the stored names when `kbInitFailed && layoutAppliedBySetup`. It's covered by *applies names corrected between a failed run and its retry*, and Local Testing on `c6459e29` confirmed it (retry with `pluginsDir=Tools` scaffolded only `Tools/.gitkeep`). `dfb39645` also corrects `isDefaultKbLayout`'s doc comment, which still said the names are applied only on the defaults.
- **`deployment-settings.service.ts:474`, `restartRequired` with the branch model applied live.** Fixed in `2a98de60`, also in `setup.routes.ts`: the route leaves out `defaultBranch` and `protectedBranches` when the save applied the branch model, just as it does for the folder names. Covered by *asks for no restart over a branch model the completing save also applied*, and confirmed in Local Testing (a completing save with branch fields returned `restartRequired: false`). `dfb39645` documents on `save()` that `restartRequired` is the service's own view and that `restartKeys` lets the caller leave out what it applied.
- **`git-root-folders.ts:143`, stat before removing `target`.** Still declined, for the reasons in the earlier reply: `target` is a fixed child of the scratch directory this call just created with `mkdtemp`, `rm` does not follow symlinks, and refusing a non-directory would only turn a harmless leftover into a failed listing.

On `dfb39645`: backend settings and startup tests 183/183, `tsc` clean in core-backend and core-frontend.

PR #201, opened 2026-09-15
fix: the path-turn test stops racing its own setup

h3xxit:
## Summary

A one-file, test-only fix for a flake that has been redding CI on unrelated branches, including the refactor stack currently in flight.

`WorkspaceService.withPathTurn > holds a delete and an upload of the same path until the turn ends` failed intermittently — most loaded full-suite runs, and about one isolated run in three — while looking fine on a quiet machine.

## It is not a product bug

A turn is claimed **asynchronously**. `withPathTurn` awaits the workspace directory before it reaches the queue in `withResolvedPathTurn`, so *making the call does not mean the turn is held yet*.

The test launched the delete and the upload immediately after that bare call. Each of those awaits the same directory lookup, so whichever resolved first claimed the turn. Usually the holder, because it was called first; under load the delete, which then ran to completion and failed the assertion with a `deleted: true` that said nothing about whether turns exclude each other.

So the test was racing its own setup, and the thing it reported was scheduling order rather than the property it exists to check.

## The fix

The held callback runs **only once the turn is held**, which makes it the signal the test actually needs. It now resolves an `inTurn` promise on entry, and the test awaits that before launching either mutator. The assertion is then about turn-taking and nothing else.

The five real `fs.stat` round trips stay: they are still what gives a hypothetically-unblocked `fs.rm` time to land, and the comment explaining why a microtask yield would prove nothing is unchanged.

## Test plan

- [x] 5 of 5 passes in isolation (previously ~2 in 3)
- [x] Full backend suite clean: 212 files, 2721 tests, 4 skipped — the run that reproduced the failure most reliably
- [x] `tsc --noEmit` clean

---
## Summary by cubic
Fixes the `WorkspaceService.withPathTurn` test that was racing its own setup and flaking CI. The turn is claimed asynchronously, so launching the delete and upload right after the `withPathTurn` call let whichever mutator resolved the directory lookup first take the turn — usually the holder, but under load the delete, which then failed the assertion. The test now awaits an `inTurn` promise that the held callback resolves on entry, so the mutators only run once the turn is actually held. This is test-only; the five real `fs.stat` round trips and the comment explaining why a microtask yield would prove nothing are unchanged.

<sup>Written for commit 5ed94a95b6be6c79786a804d7e18da0879a4d844. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Bevel-Software/Hexis/pull/201?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


h3xxit:
Superseded by #200, which landed the same fix first — closing rather than rebasing a duplicate.

Same diagnosis (a turn is claimed only after `resolveWorkspaceDir`, so calling first is not acquiring first) and the same mechanism (the held callback signals on entry, the test awaits that before launching the contenders). #200 is the better change: it fixes **both** path-turn tests, and for the delete/sync one it asserts what each callback actually saw on disk instead of asserting call order, which is the real property.

The conflict flagged here was not in that test — my branch predates #185's stateless MCP endpoint landing in `dev`, so it diffed against the whole of that too.

Nothing of mine is lost: `dev` already covers the one test I had fixed.

PR #203, opened 2026-09-16
CR-Binary-File-Preview — A change request shows a proposed document or image with the file page's viewer

h3xxit:
Ticket: `Data/Engineering/Knowledge/Hexis/Tickets/CR-Binary-File-Preview.md`

## Summary

The change-request dialog now renders an added/changed binary file (pdf, docx, pptx, xlsx, csv-diff-exempted, images, eml/msg) with the same read-only viewer the file page uses, bound to the request branch's workspace, instead of the previous "No text to show" placeholder.

- Changed files show a "Proposed version" label with an "Open the current version" link that opens the file on the target branch in a new tab.
- Untouched binary files in the dialog now show the current version with the viewer.
- Formats without a viewer (`.doc`, `.ppt`, `.xls`, `.odt`, `.odp`, `.ods`, `.zip`, unknown binaries) keep the existing note and gain a "Download the proposed file" action.
- Reading proposed bytes goes through the same access check as the file page, so a reviewer without read access sees the same refusal.
- New: `RendererWorkspaceContext` / `useRendererWorkspaceId`, `hasFileViewer`, `BranchFilePreview`/`BranchFileDownload`, a `label` prop on `DownloadFileButton`.
- Viewer chunks (Pdf/Docx/xlsx renderers) stay lazy; the text/markdown diff path is unchanged. `.csv` deliberately keeps its text diff rather than moving to the CSV viewer, to protect that diff path (`isBinaryFile` is false for csv; `hasFileViewer('.csv')` is still true for the file page itself).

## Testing

- `pnpm typecheck`: 0 errors.
- `pnpm test`: 2758 passed / 1 skipped across 213 files (includes 6 new dialog binary-pane tests plus `hasFileViewer` cases).
- eslint at existing 181-problem baseline (no new-file issues); design-system ratchet shows no regressions.
- Local from-source docker boot verified in-browser as admin and as a non-admin reviewer: proposed pdf/docx/image rendering with the request branch's bytes, "Proposed version" + open-current link, untouched-binary current-version rendering, no-viewer download fallback, and the access refusal matching the file page.
- Eager bundle: 662.04 → 663.20 KB gzip; viewer chunks unchanged and still lazy.

---
## Summary by cubic
The change-request dialog now uses the file page's read-only viewer for supported binary files instead of the "no text to compare" placeholder, letting reviewers inspect content before applying a request.

**Behavior**
- Added and changed files read from the request branch and show "Proposed version"; changed files link to the target branch's current version in a new tab.
- Renames use the previous path for that link, while renames without one omit it.
- Removed and untouched files read from the target branch and show "Current version"; removed files identify the pending deletion.
- Unsupported formats such as legacy Office, OpenDocument, `.zip`, and unknown binaries keep the note and offer a download from the relevant branch.
- Previews and downloads ask the same access check as the file page: a refused read shows the same refusal, and a read-but-not-download verdict disables the button with its reason.
- Failed image reads now say so and offer "Try again" instead of leaving the pane on "Loading image…"; `.csv` remains on the text diff path.

**Implementation**
- Added `RendererWorkspaceContext` and `useRendererWorkspaceId` so renderers can read from branch workspaces without exposing editing state.
- Added `BranchFilePreview`, `BranchFileDownload`, registry-derived `hasFileViewer`, and customizable `DownloadFileButton` labels.
- Image previews now refetch after branch changes and report HTTP or network failures; viewer chunks remain lazy.

<sup>Written for commit 38ae4ecf3b5a564bb41e80f12a7ff9e6c028267c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Bevel-Software/Hexis/pull/203?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


h3xxit:
All six findings accepted and fixed in `2ecba3a4`. Two are answered differently from the suggestion, so here is the reasoning per thread.

**`ChangeRequestDialog.tsx:305` — removed binary (P1).** Real: the request's branch does not have the path, so the viewer 404s. Fixed, but NOT by going back to the no-preview note. A deletion is the one case where the document itself is the whole question being asked, and it still exists — on the target branch. So a removed binary reads the TARGET branch, labelled "Current version", with `This request DELETES this file. Below is the version on <target> that would go.` above the pane. Its no-viewer download moved to the target branch as well, which had the same 404 (the thread only named the viewer).

**`ChangeRequestDialog.tsx:312` — renamed binary (P2).** Correct. New `currentVersionPath` prop carries `previousPath` through to the link, and a rename the request described *without* a `previousPath` now offers no link at all rather than one pointing at a path the target branch has never had.

**`BranchFilePreview.tsx:125` — default-allow + a spinning image (P2).** The permanent spinner was real; the default-allow is not the part to change (it matches `useFileAccess`, and a transient lookup failure must not manufacture a refusal). Took the other option the thread offers: `ImageRenderer` now surfaces the raw-request failure — `Couldn't load this image (HTTP 403).` — instead of returning silently on a non-ok response. That also closes the identical hole on the file page, and it is what makes the pane's default-allow honest: the viewer's own error is the fallback, so every viewer mounted here has to have one. `useBranchFileAccess`'s doc now records that dependency.

**`BranchFilePreview.tsx:234` — stale image after a branch change (P2).** Correct, and the doc comment on `useImageRevision` explicitly claimed the change-request dialog shows no live image — no longer true as of this PR. `ImageRenderer` now folds the revision for the workspace it is pointed at (here, the request's branch) into its URL, and the comment is updated.

**`BranchFilePreview.tsx:286` — silent bootstrap failure (P2).** Correct. `BranchFileDownload` handles `failed` separately from loading and says so. The preview pane's own failure wording dropped "the change request's branch", since this pane reads the target branch too (untouched and removed files).

**`.changeset:2` — patch vs minor (P2).** Agreed, now `minor`.

Tests added: removed binary reads the target branch and is named as a deletion; removed no-viewer download; renamed binary's current-version link resolves to the old path; rename without `previousPath` offers no link; bootstrap failure on the download path; `ImageRenderer` non-ok, thrown, and revision-refetch cases. Suite is green (`pnpm typecheck` clean, 4,971 tests across 385 files), eslint back at the 181-problem baseline, design-system ratchet shows no regressions, and the viewer chunks are still lazy.

h3xxit:
Head is `3ace804d`, not the `2ecba3a4` named above — amended to reattach a doc comment that had drifted off `move` onto the new `selectedFile` lookup. No behaviour change between the two.

packages/core-frontend/src/modules/change-requests/components/BranchFilePreview.tsx
```diff
@@ -0,0 +1,340 @@
...
+        // note above. What makes it safe is that the viewer this then mounts
+        // reports its own rejected read — the pane hands over to the raw
+        // endpoint's answer rather than claiming the read succeeded. Every
+        // viewer here says so out loud (`ImageRenderer` was the last one that
+        // did not, and now names the status); a viewer that swallowed a 403
+        // would strand this path on a spinner.
+        //
+        // `canDownload: null` — unknown, so the button stays clickable and the
+        // backend keeps the last word on it.
+        if (!cancelled) setAnswer({ key, access: { canRead: true, canDownload: null } });
```
h3xxit, line 137: Addressed in `3ace804d`, in the viewer. `ImageRenderer` no longer returns silently on a non-ok response — it renders `Couldn't load this image (HTTP 403).` — so this default-ALLOW hands over to the raw endpoint's answer instead of stranding the pane on `Loading image...`. Tests: *names a failed read instead of spinning on it forever* and *names a read that threw* in `ImageRenderer.test.tsx`.

I did not take the other option in the finding (a failure state for the access check). `useFileAccess` defaults to allow for the same reason — a transient lookup failure must not manufacture a refusal — and a third "lookup failed" state would change nothing a reader sees, since the bytes are gated where it counts. What the finding correctly exposed is that the default-allow *depends* on every mounted viewer having an error state; `ImageRenderer` was the last one without one. That dependency is now stated here and on the hook's doc comment.

packages/core-frontend/src/modules/change-requests/components/BranchFilePreview.tsx
```diff
@@ -0,0 +1,340 @@
...
+              same lookup, with the same identity, `FileViewer` mounts. */}
+          {/* eslint-disable-next-line react-hooks/static-components */}
+          <Renderer
+            // The viewers here fetch their own bytes; the text buffer is
+            // never read, and there is nothing to seed it from on a branch
+            // the dialog deliberately does not check out.
+            //
+            // Which also means the FRESHNESS of those bytes is the viewer's
+            // own business, and the branch under review is one its author may
+            // still be pushing to. `ImageRenderer` folds this workspace's
```
h3xxit, line 265: Addressed in `3ace804d`, in the viewer. `ImageRenderer` folds `useImageRevision(workspaceId)` into its raw URL as `&v=`, counted for the workspace it is pointed at — here the request's branch — so a replaced picture reaches this pane while it stays open. Test: *refetches when an image in the workspace changes* in `ImageRenderer.test.tsx`.

Worth noting the finding also caught a stale claim: `useImageRevision`'s own doc comment asserted that the change-request dialog shows no live image, which this PR made untrue. That is corrected.

One limit, now recorded here: a document viewer has no equivalent signal to fold in — `file-changed` carries a path and only the image resolvers count it — so a pdf or a deck replaced mid-review is still read at the revision the pane opened on.

h3xxit (commented):
Answering the two findings cubic still lists as unresolved. Both were fixed in `3ace804d`, but in `ImageRenderer.tsx` rather than at the `BranchFilePreview.tsx` lines they are anchored to — which is why they still read as open. `742a402c` names the guarantee at each anchor so a reader there is not left to find it.

No disagreement with either finding; only with the location the fix belongs in. Both defects were properties of the viewer, and fixing them at the call site would mean this pane pre-fetching bytes that every viewer already fetches for itself.

packages/core-frontend/src/modules/change-requests/components/BranchFilePreview.tsx
```diff
@@ -0,0 +1,357 @@
...
+    return (
+      <p className="text-center text-detail text-ink-faint">
+        That branch couldn't be opened, so the file can't be downloaded here.
+      </p>
+    );
+  }
+  // Nothing yet — and no button either. An enabled Download that turns out to
+  // be a refusal is worse than one that arrives a moment later.
+  if (!workspace || access.canRead === null) return null;
+  if (access.canRead === false) return <AccessRefused repoRelativePath={repoRelativePath} />;
```
h3xxit, line 342: Fixed in `38ae4ecf` — and this one was a genuine gap in the PR's own acceptance criterion, which asks that reading the proposed bytes go through the same access check as reading the file on that branch. The viewer pane did; this path asked nothing at all, so a restricted `.zip` or legacy `.doc` offered a live Download whose click returned a bare `Download failed (HTTP 403)`.

`BranchFileDownload` now runs the same `useBranchFileAccess` lookup the viewer pane runs:
- `canRead === false` → `FileRoute`'s own refusal wording, no button, and the bytes are never requested.
- `canDownload === false` → the button stays, disabled, with `You do not have download permission for this file.` — `DownloadFileButton`'s existing rule, fed the same way `FileViewer` feeds it.
- `null` (in flight) → no button yet, rather than an enabled one that may turn out to be a refusal.

Tests: *refuses the no-viewer download when the reader may not read the file* and *disables the no-viewer download when the reader may not download it*.

packages/core-frontend/src/modules/workspace/components/renderers/ImageRenderer.tsx
```diff
@@ -1,35 +1,95 @@
...
-  }, [workspaceId, filePath]);
+  }, [workspaceId, filePath, revision, attempt]);
+
+  if (read.error) {
+    return (
+      <div className="flex h-full flex-col items-center justify-center gap-3 p-4">
+        <p role="alert" className="text-detail text-danger">
+          {read.error}
+        </p>
+        <Button variant="outline" size="tiny" onClick={() => setAttempt((n) => n + 1)}>
```
h3xxit, line 85: Fixed in `38ae4ecf`. Correct that the error was terminal: a failure whose `workspaceId`, path and revision are all unchanged has nothing to re-trigger the effect, so a dropped connection or a 502 stranded the pane until it remounted.

`Try again` bumps an `attempt` counter that the read depends on, which re-runs the same request. Automatic recovery on a path or revision change is untouched — those move the deps by themselves, so the retry is additive rather than a replacement for it.

Test: *recovers from a transient failure when asked to try again* — first read rejects, the retry succeeds, and it asserts the second call is the same URL and that the alert is gone.

h3xxit (commented):
Both findings from the pass on `742a402c` accepted and fixed in `38ae4ecf`. The download one was a real gap in this PR's own access criterion, not just a rough edge — thanks for it.

On the re-detected `BranchFilePreview:137` finding (the access default-ALLOW): the code has moved under it twice now, and as of this commit the path it describes ends in a named error *with a Retry*, not a spinner. I'd argue that is better than either option the finding offers — a manufactured refusal from a transient lookup failure would be wrong (`useFileAccess` defaults to allow for exactly that reason), and the viewer reporting its own rejected read is what keeps the pane honest about whose answer it is showing. Flagging for the human reviewer rather than changing it again.

PR #204, opened 2026-09-16
Delete-Confirm-And-Move-Warning — Deleting asks first, and moving a file between folders says what changes before it happens

h3xxit:
Implements `Data/Engineering/Knowledge/Hexis/Tickets/Delete-Confirm-And-Move-Warning.md`.

## Summary
- Choosing Delete now opens a confirmation dialog: "Delete <name>?" for a file, or "Delete <folder> and its N files?" for a folder (N counted from the tree). Cancel does nothing; Confirm performs the existing delete. Reserved roots remain undeletable as before.
- Dropping a file or folder into a different folder opens a confirmation before anything is sent: "Move <name> to <destination>? Access to it will follow <destination>'s rules from now on." Confirm performs the existing move; Cancel leaves everything untouched. Dropping onto its own folder remains a no-op.
- The move confirmation adds contextual warnings when applicable:
  - "You can't write to <destination> — the move will be refused" (verified: unwritable moves are refused today, not proposed).
  - "<name> is a platform-managed file; moving it changes how the platform reads it" for `AGENTS.md`, `roles.yaml`, `access.md`.
  - "This moves it out of <root>/ into <root>/ — the two roots are handled differently" when source and destination roots differ.
- Keyboard: Enter confirms, Escape cancels; focus returns to the row afterwards.
- New `treeConfirm.ts` utility computes the confirmation copy; `TreeActionConfirm.tsx` renders the dialog; `FileExplorer.tsx` wires delete/move actions through the new confirmation flow.
- Tests cover delete confirm (file, folder with count, cancel), move confirm (base sentence, each of the three extra warnings, cancel sends nothing, same-folder drop is a no-op), and keyboard handling.

## Test plan
- `pnpm typecheck` — pass
- `pnpm test` (core-frontend vitest) — 1996/1996 pass
- Full source build against the compose app + headless Chromium manual verification of delete/move confirmation flows

---
## Summary by cubic
The sidebar used to delete or move on the first click or drop; it now asks first: Delete opens “Delete <name>?” (“Delete <folder> and its N files?” for folders), and a drop into another folder opens “Move <name> to <destination>? Access to it will follow <destination>'s rules from now on.” Nothing is sent until Confirm, and Cancel leaves everything untouched.

**Warnings**
- Move confirmations add a warning when you can't write to the destination, when the entry is a platform-managed file (`AGENTS.md`, `roles.yaml`, `access.md`), or when the move crosses reserved roots.

**Side effects**
- Moves the server refuses now surface an alert instead of failing silently.
- Reserved root folders stay undeletable, and dropping an entry onto its own folder remains a no-op.
- Enter confirms, Escape cancels. Cancel returns focus to the row; Confirm hands it to a row that stays (the drop target or the containing folder).
- An open confirmation is stamped to its workspace; switching workspaces drops it, so Confirm can't act on a different branch.

<sup>Written for commit 7d88b5526f90df805c9e6054eff17a63d1eecdca. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Bevel-Software/Hexis/pull/204?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


h3xxit:
Response to the cubic review of `bd8ca0bc`. Fixes are in `7d88b552`, which also merges `origin/dev`.

**P1: Confirm could run against the wrong workspace** (FileExplorer.tsx:1211, [thread](https://github.com/Bevel-Software/Hexis/pull/204#discussion_r4024178246)): **Fixed.** Each open request now records the workspace it was opened in. The dialog only shows while that is still the current workspace, and an effect clears the request when the workspace changes, so switching back doesn't bring it back. New test: `drops an open confirmation when the workspace changes under it`.

**P3: Focus fell to `body` after a confirmed move or delete** (FileExplorer.tsx:732, [thread](https://github.com/Bevel-Software/Hexis/pull/204#discussion_r4024178256)): **Fixed.** Each request now has a `focusAfterRun` target. After Confirm, focus goes to the row the item was dropped on (move) or to its parent folder (delete). Both rows stay in the tree after the refresh. Cancel and Escape still return focus to the original row. The Enter test now checks the drop-target row, and a new test covers a confirmed delete.

**P3: The branch model leaked out of the tests** (FileExplorer.test.tsx:1395, [thread](https://github.com/Bevel-Software/Hexis/pull/204#discussion_r4024178264)): **Fixed.** An `afterEach` restores the branch model from `test-setup.ts`, following the pattern in `bootstrap.test.ts`.

**P2: Use `useLayoutEffect` for focus restore** (FileExplorer.tsx:1216, [thread](https://github.com/Bevel-Software/Hexis/pull/204#discussion_r4024178250)): **Not changed, on purpose.** The shared `Dialog` restores focus to the element that had it before the dialog opened, and it does that in a passive `useEffect` cleanup. React runs passive cleanups after layout effects. With `useLayoutEffect`, we would focus the row first and the Dialog would then move focus back to the element that had it before the dialog opened (the closed menu, or whatever had focus before the drop). The row focus has to run as a passive effect so it comes after the Dialog's restore. I tried the change: 3 of the 4 keyboard tests failed (Enter confirm, Escape cancel, confirmed delete), so I reverted it.

Checks: `tsc --noEmit` exit 0; core-frontend vitest 1998/1998 passed.

PR #207, opened 2026-09-16
OIDC-Settings-Apply-Live — Single sign-on configured on the setup screen is available without a restart

h3xxit:
Implements `Data/Engineering/Knowledge/Hexis/Tickets/OIDC-Settings-Apply-Live.md`.

## Summary
- OIDC settings (issuer, client id, secret, scopes, label) saved via the first-run screen or the Deployment page now take effect on the next sign-in attempt, with no server restart required.
- `GET /api/auth/providers` only advertises the OIDC provider once issuer, client id, and secret are all configured; `/api/auth/oidc/login` redirects to the sign-in page with a `not_configured` error instead of a 500 while incomplete.
- Changing the issuer triggers re-discovery instead of reusing a stale discovery document; label/scopes/client id/secret changes are read live on each sign-in attempt.
- The five OIDC settings no longer report `restartRequired`; settings that still need a restart (e.g. `allowedEmailDomains`) are unaffected.
- Environment-configured `OIDC_*` values continue to take precedence over UI-saved values, and UI saves to env-backed keys are refused.
- Non-OIDC providers supplied by an overlay are unaffected (checked via `isEnabled?.() ?? true`).

## Testing
- core-backend vitest: 2847 passed; typecheck clean across packages.
- Local build from source (`docker-compose.build.yml`) against a mock HTTPS OIDC IdP: verified unconfigured state (no provider, `not_configured` redirect), button appearing after secret save without restart, live label/scopes/client id/issuer changes, issuer change re-discovery, and a full PKCE sign-in via headless Chromium.
- `oidc-auth-provider` + `deployment-settings` vitest: 36/36 passed.

---
## Summary by cubic
OIDC settings saved on the setup screen or Deployment page previously required a server restart; they now apply to the next sign-in attempt. The provider is registered once at boot and reads its configuration on every probe and sign-in.

**Behavior**
- `GET /api/auth/providers` only lists OIDC once issuer, client id, and secret are all set.
- An unconfigured login redirects to sign-in with `not_configured` instead of a 500.
- Label, scopes, client id, and secret changes apply immediately.
- The discovery document is cached per issuer and shared across concurrent sign-ins; changing the issuer — or switching back after a failed change — triggers re-discovery.
- The five OIDC settings no longer report `restartRequired`.
- Environment `OIDC_*` values still take precedence, and UI saves to env-backed keys are refused.
- Other auth providers are unaffected.
- On multi-replica deployments, a saved OIDC change reaches only the replica that served the save until the others restart.

<sup>Written for commit 392ac9fd0ef2465e848bc50f65a999c967c641d0. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Bevel-Software/Hexis/pull/207?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


h3xxit:
Replies to the cubic review of `b5c9264f`. The fixes are in commit `392ac9fd` ("fix: OIDC discovery is shared in flight and never outlives its issuer"). It goes up with the next push.

1. **Rotated secret never reaches the token exchange in tests** (https://github.com/Bevel-Software/Hexis/pull/207#discussion_r4025274785): **Accepted.** New test `exchanges the code with a rotated client secret on the next sign-in`. It runs a full login and callback, saves a new `oidcClientSecret`, runs the flow again, and checks that the token endpoint's `Authorization: Basic` header uses the new secret.

2. **Old issuer's document served after a failed change and switch-back** (https://github.com/Bevel-Software/Hexis/pull/207#discussion_r4025274791): **Accepted.** The cache holds only the issuer that was last requested. It is replaced as soon as a different issuer is requested, and dropped if that fetch fails. Switching back therefore fetches discovery again. New test: `re-discovers the original issuer when switched back after a failed change`. It fails on the old code.

3. **Concurrent discovery fetches after an issuer change** (https://github.com/Bevel-Software/Hexis/pull/207#discussion_r4025274821): **Accepted.** The cache now stores the in-flight promise, so concurrent sign-ins share one fetch. A failed fetch is still not cached. New test: `shares one discovery fetch between concurrent sign-ins`, with a slow issuer so the sign-ins overlap. It fails on the old code.

4. **Other replicas keep the old settings** (https://github.com/Bevel-Software/Hexis/pull/207#discussion_r4025274812): **Partly accepted.** The caveat is valid, so `docs/configuration.md` now says the change applies on the replica that handled the save, and that other replicas need a restart, especially after rotating the client secret. I haven't brought `restartRequired: true` back in the save response. The ticket's acceptance criteria explicitly require that the five OIDC settings no longer show the restart notice. The server also can't tell whether it is one replica of several, so the flag would be wrong for every single-replica deployment. The proper fix is to invalidate settings over the event bus, as the `DeploymentSettingsService` comment says. That's a separate piece of work and outside this ticket.

PR #208, opened 2026-09-16
Sidebar-Rows-Say-What-They-Are — A sidebar row shows whether it is an empty folder, and a long file name keeps its extension

h3xxit:
Implements `Data/Engineering/Knowledge/Hexis/Tickets/Sidebar-Rows-Say-What-They-Are.md`.

## Summary
- Every folder row now shows the caret, including a folder with no children; opening an empty folder shows one muted, control-free "Empty" row beneath it.
- A file name that does not fit is truncated in the middle instead of at the end: the leading part is shortened with an ellipsis and the last 8 characters (which include the extension for ordinary names) always stay visible. A name that already fits is unchanged.
- Folder names still truncate at the end, since they have no extension to protect.
- Row height, indent, hover/current tones, keyboard navigation and the context menu are unchanged, and no icons were introduced.
- Both the Knowledge explorer and the Skills tree pick this up automatically since they share the row component.

## Testing
- `pnpm install` + frontend typecheck (clean, after building `@bevel-software/platform-shared` and `platform-mcp-core`).
- `vitest` (NODE_ENV=test): full frontend suite 2019 passed; 89/89 in the two changed test files, covering the new AC cases (caret on an empty folder, the "Empty" row on open, middle truncation keeping the extension, a short name rendering whole, folder truncation unchanged).
- Local end-to-end pass: built from source via `docker-compose.build.yml`, verified in headless Chromium on both trees — empty folders show a caret and open to one muted, control-free Empty row; long file names cut in the middle keeping the last 8 characters; short/fitting names render whole; long folder names still cut at the end; row height, hover/current styles, and right-click menus unchanged; no new icons.

---
## Summary by cubic
Makes sidebar rows state what they are: every folder now shows its caret, opening an empty folder shows a muted "Empty" row, and long file names truncate in the middle so the extension stays visible.

- Empty folders previously had no caret; now they get one and open to an inert "Empty" row.
- File names that don't fit keep their last 8 characters on screen; folder names still truncate at the end.
- The shared row component means both the Knowledge explorer and Skills tree pick this up automatically.
- No icons introduced; row height, indent, hover/current tones, keyboard navigation, and context menu are unchanged.

<sup>Written for commit d0979230cf8ce605bfc84422660d8ca69b84b450. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Bevel-Software/Hexis/pull/208?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


h3xxit:
Re cubic's finding on `FileExplorer.tsx:909` (https://github.com/Bevel-Software/Hexis/pull/208#discussion_r4025480604) — keeping `aria-expanded={isExpanded}` on empty folders. Here's why.

The earlier team rule (95bac39a, "an empty folder claims no expanded state") was written when an empty folder had no caret and **could not open**. Clicking did nothing, so `aria-expanded` would have described a control that didn't exist. This PR changes that on purpose. The ticket says every folder row gets the caret, and opening an empty folder shows an "Empty" row beneath it. The empty folder's button now does toggle content: click it and the "Empty" row appears, click again and it goes away.

`aria-expanded` is the attribute that tells assistive tech a button shows and hides content, and what state that content is in. If we drop it only for empty folders:
- a screen reader would announce a plain button, and pressing it would silently add a line of text with no state change announced;
- sighted users would see a caret that turns while screen-reader users get no matching state, which makes the two experiences disagree.

So the reason behind the old rule no longer applies, and following it now would hurt accessibility. The tests assert this in both trees: `FileExplorer.test.tsx` ("gives a childless folder the caret", "shows one muted, inert Empty row", which checks `false` → `true` on click) and `SkillsTree.test.tsx`.

If reviewers would rather empty folders stay out of the expand/collapse model entirely, the right fix is to remove the caret and the Empty row, not only the attribute. That would be a change to the ticket's requirements, not a code fix.

PR #217, opened 2026-09-16
Copy-Path-Is-A-Link — Copy path gives a root-anchored path that works in a link, and pasting a path or URL into the editor makes a link

h3xxit:
Ticket: `Data/Engineering/Knowledge/Hexis/Tickets/Copy-Path-Is-A-Link.md`

## Summary
- Copy path (sidebar row menu and the page-level Share menu, which previously had no Copy path item) now copies a root-anchored path `/<kbDirName>/<path>` that opens the file from any folder when pasted into a Markdown link.
- Every agent/workflow tool that takes a path (`path`, `src`, `dest`, `destination`, `files[].path`) now accepts one leading slash and treats it as the same workspace path, via `normalizePathArgs` in `kb-fs/repo-path.ts`. Tool descriptions were updated to say so. The refusal for paths outside the repository is unchanged.
- Pasting into the Markdown editor a clipboard that is exactly one bare workspace path (`/<kbDirName>/...`) or one `http(s)` URL, with no selection, now inserts a Markdown link: the label is the file name without extension for a path, or the host + path for a URL. With a selection, the selection becomes the label. Any other paste is unchanged.
- Added a changeset.

## Test plan
- Backend and frontend unit tests cover: Copy path output, tools accepting the leading slash, paste with no selection (path/URL/ordinary text), and paste with a selection.
- `pnpm typecheck` passes.
- Verified locally in a docker-compose build: Copy path from sidebar and Share menu copies the root-anchored path and the pasted link opens the file from a subfolder; 11 MCP file tools accept a leading slash with descriptions updated and outside-repo refusal unchanged; editor paste correctly links paths/URLs, uses selection as label, and leaves ordinary text/sentences/`/etc/passwd` pastes unchanged; undo works.

---
## Summary by cubic
Copy path now copies a root-anchored path (`/<kbDirName>/…`) that opens the file from any folder when pasted into a Markdown link, and pasting a bare workspace path or URL into the Markdown editor inserts a link. Previously the copied path had no leading slash, so a link pasted from a subfolder resolved against that folder and ended in File not found.

**Agent tools**
- Every path tool input (`path`, `src`, `dest`, `destination`, `files[].path`) accepts one leading slash as the same workspace path via `normalizePathArgs`, and its description says so; paths outside the repository are still refused.
- `post_change_request_comment` drops the clone folder from a copied path so the comment anchors to the repository-relative file.
- The page-level Share menu gains the Copy path item it lacked before, and the workspace root row no longer offers Copy path (it would copy `/.`).

**Editor paste**
- Pasting exactly one bare workspace path or `http(s)` URL with no selection inserts a link labelled with the file name without its extension, or with the URL's host and path.
- `#` and `%` in a pasted path are percent-encoded so the link resolver doesn't read them as a fragment or decode them.
- A selection becomes the label; all other pastes, including multi-line text, are unchanged. A changeset is included.

<sup>Written for commit 0582b1a89059d69b560833061da357e6e75fec42. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Bevel-Software/Hexis/pull/217?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


h3xxit:
Replies to the cubic review on 8d2263d3. Fixes are in commit 0582b1a8, which will be pushed on the next pass.

1. **Changeset patch → minor** ([r4030046619](https://github.com/Bevel-Software/Hexis/pull/217#discussion_r4030046619)): **Fixed.** Both packages are now `minor`.
2. **`post_change_request_comment` stores the path with the clone folder** ([r4030046639](https://github.com/Bevel-Software/Hexis/pull/217#discussion_r4030046639)): **Fixed.** Change-request files are repository-relative (`git diff --name-status`), so the handler now drops a leading `<kbDirName>/` after the slash is normalised. `/knowledge-base/KnowledgeBase/Foo.md`, `knowledge-base/KnowledgeBase/Foo.md` and `KnowledgeBase/Foo.md` all anchor to `KnowledgeBase/Foo.md`. The input description now says this, and there's a new test for it.
3. **Workspace-root row copies `/.`** ([r4030046648](https://github.com/Bevel-Software/Hexis/pull/217#discussion_r4030046648)): **Fixed.** The root row no longer offers Copy path, and a test checks this.
4. **Flat tree copies `/brief.md` with no kbDirName** ([r4030046670](https://github.com/Bevel-Software/Hexis/pull/217#discussion_r4030046670)): **I disagree, so no code change.** The explorer's tree is workspace-rooted. The root is `.`, and the clone sits under it as `<kbDirName>/`, in both the split and the flat layout, because the flat fallback renders `mergedTree` from that same root. So a real row's `relativePath` always starts with `<kbDirName>/`. The `brief.md` in that test is a synthetic fixture with no clone folder. We confirmed this in the running app, where Copy path gave `/knowledge-base/…`. A workspace file outside the clone copies as `/x.md`, and it's right that paste doesn't turn that into a link, because it isn't a repository file.
5. **`useWorkspace()` throws under `RendererWorkspaceContext` alone** ([r4030046657](https://github.com/Bevel-Software/Hexis/pull/217#discussion_r4030046657)): **Changed, but the scenario doesn't happen.** On `dev`, `MarkdownRenderer` already calls `useFileNav()` (line 65), which calls `useWorkspace()`. So the renderer already required a workspace provider, and this PR didn't add that requirement. Even so, the paste handler now reads the context with `useContext(WorkspaceContext)?.kbDirName ?? null`, so it adds no requirement of its own.
6. **`#` in a file name becomes a fragment** ([r4030046664](https://github.com/Bevel-Software/Hexis/pull/217#discussion_r4030046664)): **Fixed.** A workspace-path destination now encodes `%` → `%25` and `#` → `%23`. The resolver splits at `#` and then runs `decodeURIComponent`, so both are needed. Tests cover `C#.md` and `100%20off.md`.
7. **Description test covers only `path`** ([r4030046673](https://github.com/Bevel-Software/Hexis/pull/217#discussion_r4030046673)): **Fixed.** The test now also checks `src`/`dest` on `copy_file`/`move_file` and `files[].path` on `write_files`.

Targeted tests pass: 115 backend, 126 frontend. `tsc --noEmit` passes for core-backend and core-frontend.

PR #225, opened 2026-09-17
Binary-Capability-Contract — one binary capability contract across the agent file tools

h3xxit:
Implements Data/Engineering/Knowledge/Hexis/Tickets/Binary-Capability-Contract.md.

## Summary
- Every file tool now states the same rule for binary content in its description and in `tools_info`: `read` returns text for text files and extractions for documents; `write`, `write_files` and `edit` accept text only; `copy`, `move`, `delete` and `unzip` act on bytes of any kind; new binary content arrives through upload.
- Refusing binary content on a text tool now returns HTTP 415 with `{ kind: 'binary_not_writable', fileKind, useInstead: ['upload', 'copy_file', 'move_file'] }`, naming the file's kind and the tool to use instead. The typed refusal fields reach MCP callers via `describeToolFailure`.
- `file_stat` reports `contentMode: text | document | binary` so an agent can decide before acting.
- No tool coerces bytes through a text path; a refused batch write writes nothing.

## Test plan
- Unit: core-backend and mcp-core suites pass (typecheck + vitest), eslint clean on changed files.
- Full-source local run (docker-compose.build.yml): verified write_file/edit_file/write_files reject docx/png/zip/binary content with the typed 415 refusal; text writes/edits succeed; copy/move/unzip/upload preserve bytes exactly (sha256-checked); file_stat.contentMode correct across text/document/binary; all 12 file tools carry the content rule in tools/list and tools_info.

---
## Summary by cubic
Unifies binary handling across the agent file tools so every file tool follows one content rule. Previously only office documents were refused with HTTP 400 and text could still be written to image or archive names; now `read_file` returns text or document extractions, `write_file`/`write_files`/`edit_file` accept text only, `copy_file`/`move_file`/`delete_file`/`unzip` act on bytes, and new binary content arrives through upload. Refusals now return HTTP 415 with a machine-readable `binary_not_writable` body, which is a breaking change for callers that matched on the old 400.

**Behavior changes**
- All 12 file tools carry the same content rule in their descriptions and in `tools_info`.
- `file_stat` reports `contentMode` (`text`, `document`, or `binary`) so agents can decide before acting.
- MCP error text keeps the typed refusal fields (`kind`, `fileKind`, `useInstead`).
- A refused `write_files` batch writes nothing.

<sup>Written for commit 2375d405b302ede1c8e62b098de297c429730fb4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Bevel-Software/Hexis/pull/225?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


h3xxit:
Replies to cubic's review of `062ebe41`. Fixes are in commit `2375d405`, which will be pushed on the next pass along with a merge of `origin/dev`.

1. **`packages/mcp-core/src/results.ts` — `...details` spread runs outside the guard (P2): accepted, fixed.** Only `.error` is read outside a guard now, and that read has its own try. The copy of the details, the `kind` check and the serialisation all run under try/catch, so a throwing getter or Proxy falls back to the plain message. Details that can't be turned into plain JSON (a BigInt, a cycle) now go through `safeJsonText`, so `kind` is kept. New tests cover the BigInt case, a throwing `kind` getter and a Proxy whose `ownKeys` throws.

2. **`workspace.tools.ts` — `file_stat` reads the whole file (P2): partly disagree. Documented, not changed.** A check of only the first N bytes would be cheaper, but it would give the wrong answer. The write check refuses a file with invalid UTF-8 or a NUL byte *anywhere* in it. If `file_stat` looked only at the start of the file, it could report `contentMode: text` for a file that `write_file` then refuses. That breaks the promise `contentMode` exists for: telling an agent what will happen before it acts. The cost is also limited: only unregistered extensions trigger a read (documents, images and archives are decided by name), and it's the same single read `write_file`/`edit_file` already do on that file. The filesystem layer (Mastra `LocalFilesystem`) has no partial-read API to build a cheaper exact check on. The trade-off is now written in a comment next to the read.

3. **`workspace.tools.test.ts` — `not.toBe(415)` also passes when the batch fails (P3): accepted, fixed.** The test gives the plain filesystem a `writeFiles` that writes each file, so the text-file `write_files` call must now return `200`.

Also fixed: the refusal message read "this is a image file" / "a archive file" (flagged in local testing). It now reads "this file's kind is image".

PR #230, opened 2026-09-17
Empty-Workspace-Says-Why — an empty tree says why — nothing shared vs. empty knowledge base

h3xxit:
Implements `Data/Engineering/Knowledge/Hexis/Tickets/Empty-Workspace-Says-Why.md`.

A workspace with nothing the user may read now says so instead of showing a blank tree.

## Summary
- The files listing root (`GET /workspace/:id/files`) now carries an optional `withheld` count: the number of entries hidden from this user by permissions. `.bevelignore`-hidden entries are not counted (they're "unlisted", not withheld) and admins never see the key.
- A new `EmptyTreeNotice` renders in `FileExplorer.tsx` and is reused by the Skills/Plugins trees (`RootFolderTree`), since they share the same filtered listing:
  - `withheld > 0` → "Nothing here is shared with you yet. Ask an admin to grant you access."
  - otherwise (KB has no files at all) → "This knowledge base is empty.", followed by the create hint when the user can write at the root.
- Neither message renders once at least one entry is visible.

## Tests
- Backend: `withheld` is computed only from filtered-out keys, not `.bevelignore`-hidden ones.
- Frontend: both empty-state messages, and confirmation that neither renders with one visible entry.

core-frontend 2087 passed, core-backend 2965 passed / 1 skipped, both typecheck clean. Verified locally against a source build (see Local Testing step) covering both messages, the withheld count/no-count cases, the create hint, and the Skills tree reuse.

---
## Summary by cubic
When a workspace tree has nothing visible, it now explains why. The file listing root carries a `withheld` count of entries hidden by read rules, and the Knowledge explorer and Skills/Plugins trees show "Nothing here is shared with you yet" when that count is non-zero, or "This knowledge base is empty." otherwise.

- `withheld` appears only when non-zero and carries only a number, never names.
- The admin-only `.bevelignore` is hidden as `unlisted` and does not count as withheld.
- The empty message adds a create hint only when the caller may write at the tree's root folder; on trees that predate the Knowledge split, that check targets the KB clone folder itself.
- Neither message renders while the tree loads or once any entry is visible.

<sup>Written for commit bdf0ff9cb6d0aa0d0a3bbabfe5e068e720eb75ee. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Bevel-Software/Hexis/pull/230?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


h3xxit:
Addressed cubic's review of 4d8707e4 in bdf0ff9c (branch also synced with `dev`):

1. **read-filter test, line 178 (`toBeGreaterThan(0)`)**: accepted. The all-denied case now asserts the exact count: every entry in the unfiltered tree except the root.
2. **FileExplorer.tsx, line 1694 (legacy tree checks `.`)**: accepted. When there's no split and the tree wraps the KB clone's folder, the notice now uses that folder as its root. `useCanWriteFolder` treats a bare `kbDirName` as inside the KB and sends it unchanged; the access route's `toRepoRelative` maps that to the repo root. Before this, the `.` path skipped the lookup and always showed the create hint. New test: *asks about the KB clone folder, not the workspace root, for a tree that predates the split*.
3. **FileExplorer.test.tsx, line 1662 (test name says admin)**: accepted. Renamed to *still says "empty" when the only root entry is .bevelignore*.

Tests: FileExplorer and library tests 602/602, backend read-filter 12/12, core-frontend typecheck clean.

PR #224, opened 2026-09-17
Approve-Control-Is-Labelled — Label the approve control and summarise the viewer's approvals

h3xxit:
Ticket: `Data/Engineering/Knowledge/Hexis/Tickets/Approve-Control-Is-Labelled.md`

A reviewer sees a labelled Approve button on the file they are looking at, and the footer says what is theirs to approve.

## Summary
- The selected file's header now ends with a button: "Approve this file" when the viewer can approve and hasn't, "Approved – Undo" once they have, and nothing when they cannot approve it. Pressing it drives the same action the tree's check performs today.
- The tree's check is now a status mark only, with accessible name "Approved by you" or "Waiting on your approval".
- The footer, above the Apply button, reads "Your approval is needed on N files" with an "Approve all mine" button when N > 1 (approved sequentially), or "Waiting on <names>" when nothing is the viewer's to approve.
- The header button sits in the tab order immediately after the header text.

## Testing
- `pnpm typecheck` clean; core-frontend vitest 156 files / 2062 tests pass (change-requests: 8 files / 67 tests, including the four required cases: one file pending, two files pending, none pending, and undo).
- Manually verified in headless Chromium against a real scratch knowledge base on staging, exercising approve, undo, and all footer states.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
## Summary by cubic
Implements `Approve-Control-Is-Labelled` so the approve action is findable in the change request dialog. The unlabelled green check in the file tree becomes a status-only mark, the selected file's header gets a labelled `Approve this file` / `Approved – Undo` button, and the footer reads `Your approval is needed on N files` or `Waiting on <names>`.

- The tree's status mark uses the accessible names `Approved by you` and `Waiting on your approval`; files outside the merge gate get no status mark and name nobody.
- `Approve all mine` appears when more than one file needs the viewer; it approves sequentially and surfaces one error banner if any fails.
- The header button follows the header text in tab order, and no button is rendered when the viewer cannot approve the selected file.
- Only merge-gate-relevant files count in the footer, waiting names, and the apply pass; the backend stamps each file's `inMergeGate`, and clients read that verdict instead of re-deriving the rule, so ungated files are no longer approved on the way to a merge.
- The approve controls stand down while an apply is running, and Apply stands down while an approve is.

<sup>Written for commit 73e0ab684065c543dd7312937df8ece57e32699f. Summary will update on new commits.</sup>

[...]

PR #231, opened 2026-09-17
external-tools-survive-restart — Externally registered tools and their sign-ins survive a server restart

h3xxit:
Ticket: `Data/Engineering/Knowledge/Hexis/Tickets/External-Tools-Survive-Restart.md`

## Reproduction
Restart loses nothing on staging: `list_tools`, `list_tool_setup`, and vault rows are identical before and after. The tester's original losses were an Admin-role removal (HubSpot) and a context7 sign-in that was never completed — not a restart-survival bug in the mechanism the ticket suspected.

## Fix
`list_tool_setup` now accepts an optional `branch` and returns `onBranchOnly` plus a `note` for tools declared only on that draft (not yet on the default branch, so not served). The tool description and the knowledge-base agent guide were updated to state tools are served from the default branch only, and the in-app agent defaults to its focused branch. A new test rebuilds the catalog in a fresh process over the same files and vault rows, confirming both an auto-registered and an owner-declared OAuth server stay listed and signed in after a restart, with no new client registration.

## Testing
From-source compose build at 6d201b98 (health sha matches). Build + typecheck pass; core-backend 2966 tests, core-frontend 2073 pass. Verified: merged server authorized with tools listed; draft-only server reported via `onBranchOnly` with note. After both a container restart and a recreate with the workspaces volume wiped, `list_tools`, `list_tool_setup`, and all vault rows (id/client/len/md5/updated_at) were identical.

---
## Summary by cubic
Adds restart-survival coverage for externally registered tools and their OAuth sign-ins, and makes draft-only tool declarations visible to the agent.

Reproduction on staging showed the reported losses came from an Admin-role removal and an uncompleted sign-in, not a restart bug.

- `list_tool_setup` takes an optional `branch` and returns `onBranchOnly` plus a `note` naming tools declared on that draft that the default-branch catalog doesn't serve yet; the in-app agent defaults to its focused branch, and an explicit `branch` overrides it.
- The tool description and knowledge-base agent guide now state tools are served from the default branch only, so a draft declaration stays unlisted and has no sign-in until merged.
- Draft lookups only answer for branches this process already holds a clone of, so a caller-supplied branch name can't trigger a bootstrap before the read gate runs.
- A new test drives the real `list_tool_setup` route over HTTP in a fresh process, confirming auto-registered and owner-declared OAuth servers keep their sign-ins after a restart with no new client registration.
- Fixes the `FileChangeBoxes` test fixture to mark the owned file `inMergeGate` so Apply approves it.

<sup>Written for commit f3cb84ebe5e334eca2a9059d157fc10f23563296. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Bevel-Software/Hexis/pull/231?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


h3xxit:
Both cubic findings on `6d201b98` are accepted and fixed in `030c622b`, which goes up with the next head:

1. **`tool-manuals.service.ts:211`: a caller-supplied branch was bootstrapped before the access check ran.** Valid. `listDeclaredOnlyOnBranch` now checks `workspaceService.hasBootstrappedWorkspace(workspaceIdForBranch(branch))` first and returns `[]` for any branch this process has no clone of. A made-up, private or nonexistent branch name now causes no clone and no fetch. The branch the agent drafted on already has a clone because the agent wrote to it, and after a restart the check finds that clone on disk. New test: `never bootstraps a branch this process holds no clone of`. It fails when the guard is removed.

2. **`tool-manuals.restart.test.ts:162`: the test re-implemented `list_tool_setup` instead of calling it.** Valid. Each simulated process now registers the real route through `registerToolManualsTools`, using the production tool-auth middleware, context resolver and handler factory, and calls `POST /api/agent/tools/list_tool_setup` over HTTP. The test also covers a server declared only on a draft (the tester's case): it must appear in `onBranchOnly` both before and after the restart.

The `tool-manuals` suite passes (115 tests) and the `core-backend` typecheck is clean.

h3xxit:
Re: the architecture review's **CI failing** (`test`: FileChangeBoxes). The failure did not come from this PR's changes. It also fails on `dev`:

- Since #224 (`73e0ab68`), `useApplyChangeRequest` records an approval only on files whose `inMergeGate` is true.
- The test `applies an author-approver's own proposal like the dialog's Apply` was merged separately, and its sample approval never sets `inMergeGate`. Apply therefore skips the only file, and `approvePrFile` is never called.

Fixed in `9c57ebd6`: the sample approval now sets `inMergeGate: true` (an owned markdown file). No production code changed. The branch is also merged with the latest `dev` (`fc4bb7b2`).

Locally at that head:
- core-frontend: 157 files, 2095 tests passing; `tsc` clean.
- core-backend: `tsc` clean; the tool-manuals and workspace suites pass (623 tests).

cubic's two findings were already marked addressed in `030c622`.

h3xxit:
Re: the architecture review on `9c57ebd6` (**no test for the focused-branch default**). Agreed. Added in the next commit, in `tool-manuals.tools.test.ts`, over the real tool-auth middleware and handler:

- **`defaults to the in-app agent's focused branch when no branch is given`**: an internal agent token minted with `focusedBranch: 'alice/add-crm'` calls `list_tool_setup` with no `branch`. The test asserts `listDeclaredOnlyOnBranch(alice, 'alice/add-crm')` is called, and that `onBranchOnly` and `note` name that branch. When the handler is changed to ignore `ctx.focusedBranch`, this test fails.
- **`prefers an explicit branch over the focused one`**: the token is focused on `alice/other-draft` and the call passes `branch: 'alice/add-crm'`. The lookup runs once, with the explicit branch.

The tool-manuals route tests pass (6 of 6) and the core-backend typecheck is clean. On `9c57ebd6`, the `test` and cubic checks were both green.

PR #237, opened 2026-09-17
Git-Internals-Never-Reachable — The repository's internal git folder is unreachable through every file tool and route

h3xxit:
Ticket: `Data/Engineering/Knowledge/Hexis/Tickets/Git-Internals-Never-Reachable.md`

## Summary
- Every workspace tool (stat, read, grep, list, write, write_files, edit, delete, mkdir, move, copy, unzip) and every workspace HTTP route now refuse any path whose canonical form has `.git` as a segment, in any letter case, after resolving `..`, percent-encoding, and symlinks.
- The refusal is a single sanitized message, "That path is inside the repository's internal git data and is not available.", with HTTP 403, and does not reveal whether the path exists.
- A symlink inside the repository that points into the `.git` folder is refused the same way — the check runs on the resolved real path.
- Directory listings continue to hide the git folder (and now also hide links into it).
- Archive entries inside uploaded zips that target the git folder are skipped on unzip.

## Test plan
- Backend typecheck clean.
- Full core-backend vitest run: 234 files, 3475 passed, 1 skipped — includes 466 security regression tests covering every tool and route in plain, dotted, encoded, upper-case and symlinked forms, plus 31 unit tests for the check.
- Local Docker build from source, exercised over real HTTP against a real `WorkspaceService`/`LockingFilesystem`: 551/551 probe requests across tools and routes returned 403 with the exact sanitized message, for plain, leading-slash, dotted, `./`, `%2egit`, `%252Egit`, `.GIT`, `.Git`, backslash, trailing-dot, symlinked-folder, chained-symlink, and symlinked-file forms. Existing and missing paths produced identical responses. `list_files`/`GET /files` showed no `.git` and no links into it. Unzip skipped `.git`/`.GIT` archive entries. `.git` contents were unchanged after the run.

---
## Summary by cubic
Makes the repository's internal `.git` folder unreachable through every workspace file tool and HTTP route, regardless of how the path is spelled. Any form that resolves into `.git` — any case, `..` traversal, percent-encoding, or a symlink into the folder — is refused with one sanitized 403, with an identical response whether or not the path exists.

- Refusals are enforced at the tool mount, a route guard mounted on the whole `/workspace/:id` prefix ahead of the body parser, JWT check, and every router under it, the `WorkspaceService` file methods, and the agent filesystems, before any gate, lock, or disk access; the tool-mount check never clones a branch that isn't cloned yet, and a request naming the folder in its query with a malformed body gets the 403 rather than a parser 400.
- Listings, folder downloads, and the file tree hide `.git` in any spelling and links into it; recursive listings leave the folder out instead of failing, folder deletes skip it, whole-tree greps never match inside it, and unzip skips archive entries aimed at it.
- A dangling symlink into the folder is refused too; the resolver treats only absence and link loops as unresolvable, so a permission error is thrown rather than guessed past.
- The diff ledger skips the folder on the backup side, so a legacy ledger can't read it back out through a review session, and template seeding skips entries that link into a git folder.
- Paths like `.github/` and `.gitkeep` are unaffected.

<sup>Written for commit c3169b7d37b7cac8d544fc4b3147f50dd9c39f68. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Bevel-Software/Hexis/pull/237?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


h3xxit:
**About commit c42a64fe (the `FileChangeBoxes.test.tsx` fixture)**

This commit isn't part of the `.git` guard, and it adds nothing to what this PR changes.

- **Why it was committed:** on `e8e691ba`, CI `test` failed at `pnpm typecheck` with TS1117 in `packages/core-frontend/src/modules/change-requests/__tests__/FileChangeBoxes.test.tsx:106`. The #231 merge had left `inMergeGate: true` twice in one object, and `dev` failed the same check. The PR couldn't show green CI without removing that duplicate key, so c42a64fe removed it.
- **Why it's no longer a change here:** `dev` has since received the identical fix (3cf4fe6b), and `dev` is merged into this branch. `git diff origin/dev...HEAD -- packages/core-frontend` is empty. The PR's diff against `dev` is only the 11 backend and changeset files for the `.git` guard. The commit stays in the history because the branch is merged, never rebased.
- **CI:** `test` passes on the current head (a732320b).

Nothing to review in it for this ticket.

h3xxit:
Manual review at `324c1fb2` (Razvan's side), for the agent to take over. Every item was verified against the code. The guard itself is sound: one rule (`hasGitInternalsSegment` + the resolved `assertNotGitInternals`) refuses any `.git` segment in any case, with trailing dots, percent-encoded, dotted or through a symbolic link, with one 403 whether or not the path exists; it is enforced before every file tool, in `GitGuardedFilesystem` under every agent filesystem, and in every `WorkspaceService` path method. The `readdir` override matches `LocalFilesystem.readdir` in @mastra/core 1.35.0 exactly (depth default 100, `dir/child` naming, the extension filter keeps directories). All six cubic threads are fixed and resolved. The clone's `.git/config` holds no token: the credential helper reads `$GITHUB_TOKEN` at call time.

## Fix

1. **The review router serves `.git` files.** The route guard is registered with `router.use('/workspace/:id', …)` inside `createWorkspaceRoutes`, so it only sees that router's requests. `createDiffRoutes` is mounted separately under the same `/api/workspace/:id` prefix, and its file route reads by path:

   ```
   GET /api/workspace/:id/review/file?path=knowledge-base/.git/config
   ```

   `diff.routes.ts` checks `canRead` on `.git/config` (decided by the root `access.md`, usually granted), then `DiffService.fileDiff` → `resolvePair` (containment only, `assertWithinDirectory`) → `fs.readFile(fileAbs)` → returns the text as `current`. There is no `isDiffable` gate on `fileDiff`, so `.git/config`, `.git/logs/HEAD` (every author name and email on that clone), `packed-refs`, `FETCH_HEAD` and the hook samples all come back. `acceptOne` / `rejectOne` are safe only by accident: `isDiffable` limits them to `.md`. `git-internals.security.test.ts` mounts only `createWorkspaceRoutes`, so the suite cannot see this. Verified by reading the code, not by a live request.

   Fix, three parts:
   - call `assertNotGitInternals(workspaceDir, relativePath, fileAbs)` in `DiffService.resolvePair`, which covers `fileDiff`, `acceptOne`, `rejectOne`, `syncFromDisk` and `markUserDeleted` in one place;
   - lift the route guard to the app level (`app.use('/api/workspace/:id', …)` before the routers), so every router under that prefix is covered, including any added later;
   - add the review routes (`GET /review/file`, `POST /review/accept`, `POST /review/reject`) to the route table in the security suite.

## Notes (no change needed, or follow-up)

- **Exact-spelling skips remain in four walkers:** `diff.service.ts:404` (review ledger), `tool-manuals.routes.ts:149`, `startup/steps/seed-tree.ts:99` and `startup/kb-startup-runner.ts:797`. Git itself refuses case variants of `.git`, so this is drift rather than a hole, but it is the same drift the architecture review already flagged for `enumerateFilesUnder`. Switching them to `hasGitInternalsSegment` keeps one rule.
- **The checks run twice per call.** A route with a path resolves it in the middleware and again in the `WorkspaceService` method; `LockingFilesystem` checks and then its `super` call in `GitGuardedFilesystem` checks again. Each check is two `realpath` resolutions, the root's included. Small, but it is on every read and write.
- **The architecture review check still shows failed.** Its last run said go-ahead but did not write its one-pager, so it needs `/bevel-review` triggered again.

---

h3xxit:
Picked up the manual review at `324c1fb2`. All of it is addressed; details below.

## Fix 1 — the review router served `.git` files

Confirmed, and fixed in all three parts.

- **The guard is lifted to the app.** It now lives in `modules/workspace/git-internals.middleware.ts` as `createGitInternalsRouteGuard`, and `create-core-server.ts` mounts it once for the whole prefix, ahead of every router under it:

  ```ts
  app.use('/api/workspace/:id', core.authMiddleware, createGitInternalsRouteGuard(core.workspaceService));
  ```

  That covers the review routes, and the workflow ones (`compare-file`, `show-file`, `file-at-change`), and anything mounted under the prefix later. `createWorkspaceRoutes` still mounts the same guard on its own router, so it carries the rule wherever it is mounted alone — including in the tests.
- **`DiffService.resolvePair` refuses the folder**, which covers `fileDiff`, `acceptOne`, `rejectOne`, `syncFromDisk` and `markUserDeleted` in one place. Two adjustments beyond the suggestion: the check runs *before* the two `assertWithinDirectory` calls, so a leading-slash spelling gets the one sanitized 403 rather than a containment error; and `acceptOneUnlocked` / `rejectOneUnlocked` now resolve before the `isDiffable` early return, so their safety no longer rests on nothing in the folder being markdown.
- **The security suite mounts what the server mounts:** the app-level guard, `createWorkspaceRoutes` and `createDiffRoutes`. `GET /review/file`, `POST /review/accept` and `POST /review/reject` are in the route table, so every spelling runs against them, and a new `DiffService refuses the git folder on its own` block covers the five service methods directly. Reverting `diff.service.ts` alone fails 55 of them.

## Notes

- **Exact-spelling skips:** all four now use `hasGitInternalsSegment` — `diff.service.ts` (review ledger), `tool-manuals.routes.ts`, `startup/steps/seed-tree.ts` and `startup/steps/template-files.step.ts` (which had its own `dir.toLowerCase() === '.git'`). Together with `enumerateFilesUnder`, `removeEmptyDirs` and the grep walk in the previous commit, one rule now decides everywhere.
- **The checks run twice per call:** left as is. Each layer is reachable on its own — a tool calls the filesystem without passing a route, and the app-level guard is what keeps the refusal ahead of the read gate — so dropping either turns a guarantee into an assumption about call order. Worth a follow-up if the two `realpath` resolutions ever show up in a profile; they aren't a correctness problem.
- **The architecture review check:** its last run said go-ahead but did not write its one-pager, so it still needs `/bevel-review` re-triggered.

**Tests** (`NODE_ENV=test`): build, typecheck and `ds:check` all pass; core-backend 235 files / 3600 passed, 1 skipped; core-frontend 159 / 2135; the other two packages 147 and 111.

h3xxit:
cubic's five findings on `adc1e059` are all valid and all fixed. Each fix has a test that fails without it.

- **P1, extension routes bypass the guard** (`create-core-server.ts`). The guard now mounts *before* `ext.postTools`, so an overlay surface mounted there under `/api/workspace/:id` is covered too. That mount carries no auth in front of it, so the lexical rule needed to stop depending on `req.userId` — it reads the caller's own string and touches no disk, so it answers nothing about the workspace. The resolved half still waits for the JWT check, which the second mount (after `authMiddleware`) does.
- **P1, backup ledger** (`diff.service.ts`). `walkAll` over the backup side had no skip, so a ledger seeded before this rule could still carry `.git`/`.GIT` entries; `computePending` then paired one with the workspace path and read the folder back out through the review session. It now skips by `hasGitInternalsSegment`, so legacy entries are ignored.
- **P1, unguarded body fields** (`git-internals.middleware.ts`). `paths` (the access batch) and `ancestor` (remove-from-parent) are judged like every other path-bearing field.
- **P2, `revertPlan`** (`diff.service.ts`). It resolves each path before the `isDiffable` skip now, like `acceptOne` / `rejectOne`, and it is in the `DiffService` matrix.
- **P2, template symlink** (`seed-tree.ts`). A template entry that links into a git folder is skipped: `onOther` resolves the target through `assertNotGitInternals` before `statOf`/`seedFile`, so a template that is itself a working tree cannot seed `.git/config` into a KB.

**New tests:** a `the route guard on its own` block — every path-bearing field refused before any route runs, an unauthenticated caller refused on every spelling that names the folder (the symlinked ones fall to their route's own 401, by design), an ordinary path let through; a legacy `.GIT` backup entry never surfacing in a review session; `revertPlan` in the service matrix; and a template link into `.git` not being seeded. Reverting `diff.service.ts`, the middleware or `seed-tree.ts` alone fails 11, 3 and 1 of them.

**Tests** (`NODE_ENV=test`): typecheck and `ds:check` pass; core-backend 235 files / 3621 passed, 1 skipped; core-frontend 159 / 2135; the other two packages 147 and 111.

The branch is also merged up to current `dev`.

h3xxit:
Two threads were open on `67adc824`. One is fixed, one is already fixed and the thread is stale.

**P2, malformed body preempts the refusal** (`create-core-server.ts:498`) — valid, fixed. A request naming the git folder in its query with a body that does not parse was answered 400 by the global JSON parser, before any guard ran. The guard is now mounted a third time, **ahead of the parser**, so that request gets the sanitized 403. Only the query can be judged that early, which is the honest limit: a body that does not parse names nothing anyone can read, so a `.git` path carried *only* in a broken body is still the parser's 400 — that answer says nothing about the path or whether it exists. The two later mounts are unchanged: one after parsing (body fields: `path`, `oldPath`, `newPath`, `destination`, `paths`, `ancestor`), one after authentication (the resolved form, for links into the folder).

Two tests cover it: a `.git` query with a malformed body returns the sanitized 403 and never reaches the route; the same malformed body on an ordinary path still returns the parser's 400.

**P2, `revertPlan`** (`git-internals.security.test.ts:611`) — already done in `67adc824`, the commit this thread is pinned to. `revertPlan` calls `resolvePair` before the `isDiffable` skip:

```ts
for (const rel of paths) {
  // Resolved BEFORE the diffable test, like accept and reject: the git
  // folder is refused for every path the caller names.
  const { backupAbs } = await this.resolvePair(workspaceId, rel);
  if (!isDiffable(rel)) continue;
```

and it is in the `DiffService` matrix (`['revertPlan', (p) => diffService.revertPlan(WS, [p])]`), so all 12 spellings run against it. cubic's own check on that head reported 1 issue, not 2, so I read this thread as re-anchored rather than re-raised. No code change.

**Tests** (`NODE_ENV=test`): build, typecheck and `ds:check` pass; core-backend 235 files / 3623 passed, 1 skipped; core-frontend 159 / 2135; the other two packages 147 and 111.

h3xxit:
On the CI failure at `ab194787`, and the state of the two threads.

**The `test` check failed on an unrelated flaky frontend test, not on this branch's code.** The job got through install, `notices:check`, `build` and `typecheck`, and failed in `pnpm -r run test` at:

```
packages/core-frontend/src/modules/library/__tests__/owner-pill.test.tsx:204
AssertionError: expected undefined to be true
```

That is `expect(byName.get(name)?.canWrite).toBe(true)` — an item missing from the loaded library, not a wrong value. The cause is in the test's own helper: `renderGallery` returns as soon as the cards list appears (`await screen.findByRole('list', { name: 'cards' })`) and then reads the last `onLib` snapshot, so a slice that resolves after the first paint can be missed under load.

Why it is not this branch:

- This PR touches no frontend file. `git diff origin/dev...HEAD -- packages/core-frontend` is empty.
- The file's last change is `b78e78b8`, on `dev`, present identically on both sides.
- `dev`'s own `test` check is green on that same file — the race just lost on our run.
- Locally at the merged head it does not reproduce: the file alone 3/3, and the **whole** frontend suite twice, 161 files / 2155 tests, green both times.

I have not touched that file. It belongs to the library work, and editing an unrelated test from this PR is what drew the "not tied to the stated intent" question earlier. The branch is now merged up to `dev` (`7215fd6a`), so the next push gives CI a fresh head to run.

**Threads:** cubic's check on `ab194787` reports 0 issues. The one still-open inline thread is the `revertPlan` P2, which was already fixed in `67adc824` (`resolvePair` before the `isDiffable` skip, and `revertPlan` in the `DiffService` matrix) and answered above.

**At the merged head** (`NODE_ENV=test`): build and typecheck pass; core-backend 235 files / 3623 passed, 1 skipped; core-frontend 161 / 2155.

PR #244, opened 2026-09-17
Approval-Gate-Every-File-Type — Owner approval gates a change request for every file type

h3xxit:
Ticket: Data/Engineering/Knowledge/Hexis/Tickets/Approval-Gate-Every-File-Type.md

## Summary
- The merge gate now treats every touched file the same regardless of extension (or lack of one), including binary files: any file with an eligible approver that lacks approval is reported as a missing approval.
- `mergeableInBevel` stays false while any such missing approval remains, for every file type, and missing approvals are surfaced as blocking reasons on the request detail.
- The admin bypass is unchanged: an admin can still merge past missing approvals with the bypass flag, and the warnings list what was bypassed.
- Updated the MCP tool descriptions and the change request dialog copy so neither implies the gate only applies to Markdown files.
- Added test coverage for extensionless, binary, Markdown, and mixed-file requests, with and without approvals and with the bypass.

## Test plan
- `pnpm typecheck` and `pnpm test` across shared, core-backend, core-frontend, mcp-core, and hexis-mcp.
- Built the branch from source via `docker-compose.build.yml` and manually verified: extensionless/binary/markdown/mixed requests are blocked pending approval, approving unblocks them, partial approval on a mixed request leaves only the unapproved files blocking, a non-admin bypass is refused, and an admin bypass merges with the bypassed files listed in the merge commit.

---
## Summary by cubic
Extends the change request merge gate so every touched file with an eligible approver needs approval, regardless of file type. Previously only Markdown and access-config files were gated and their missing approvals were soft warnings, so a request touching only binary or extensionless files could merge without any approval.

- A missing approval on any file with an eligible approver is now a blocking reason and keeps `mergeableInBevel` false until it's resolved.
- The admin bypass is unchanged: it merges past missing approvals and records the bypassed files in the merge commit.
- Files with no eligible approver are still ignored, so the gate can't deadlock on an unapprovable file.
- Frontend copy and the agent template no longer claim the gate only applies to Markdown; since the `merge_change_request` tool is retired on dev, its description was left as dev's.
- Fixes a regression where re-evaluating the gate after an apply failure crashed, so a recorded approval could not clear the merge block.

<sup>Written for commit 0412ec14fd885ba30bf57c6e584948d6e16e6b9d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Bevel-Software/Hexis/pull/244?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


h3xxit:
Both cubic findings accepted and fixed in `846435f4`, along with a break that merging `dev` exposed.

**1. `AGENTS.md:268` — "every file with an owner"** — agreed, and the wording was genuinely wrong, not just loose. The gate's condition is `isGateRelevant`, which is true when a file has any eligible *approver* (`eligibleApprovers.roles` or `.users` non-empty), and eligibility comes from `canWriteAtRef` — so a file with a write grant and no owner grant is gated. The old phrasing invited an agent to treat write-only files as exempt. Now reads "every affected file with an eligible approver".

**2. `approval-states.test.ts:711` — partly-approved mix** — agreed. The test only asserted `Knowledge/Foo.md` was absent, so a regression dropping one of the two *unapproved* bypass reasons would still have passed. Both lines are now asserted present.

**3. Found while syncing with `dev`:** this branch had split the gate onto a private `this.evaluateGateParts`. `dev` meanwhile added `apply-failure.test.ts`, which borrows the real gate off the prototype against a bare object (`ReviewWorkflowService.prototype.evaluateMergeGate.call({}, input)`) to get a verdict without standing up a service. After the merge that threw `this.evaluateGateParts is not a function`, and the caller swallowed it as "could not re-evaluate the merge gate; keeping its refusal" — so an approval silently failed to clear a gate refusal. Six tests across `apply-failure` and `workflow.service.folder-proposals` were failing on the merge result.

`evaluateGateParts` is now a module-level function, which makes `evaluateMergeGate` pure in `this` again; a comment on it records why it must stay that way.

**Verification** — `origin/dev` merged in (merge `7b0df415`, no conflicts). Typecheck clean across `shared`, `mcp-core`, `hexis-mcp`, `core-backend`, `core-frontend`. core-backend 262/262 files, 4244 passed / 1 skipped; core-frontend 183/183 files, 2488 passed.

PR #247, opened 2026-09-17
Delete-Tool-From-Tool-Page — An owner deletes a tool from its tool page

h3xxit:
Implements `Data/Engineering/Knowledge/Hexis/Tickets/Delete-Tool-From-Tool-Page.md`: an owner can delete a tool directly from its tool page after seeing what depends on it.

## Summary
- The tool page shows Delete to owners of the tool's plugin, in the same place the plugin page shows it (reuses the plugin page's `PageActions` component).
- The confirmation dialog lists skills whose allowed tools name this tool, plugins that link it, and the counts of stored keys and sign-ins under its name (no values); it requires typing the tool's name or an explicit confirm, matching the plugin dialog's pattern.
- Confirming deletes the `.tool` manual or the MCP server entry, wipes stored secrets under that name, and returns to the plugin page; the library reloads so the tool disappears from cards, sidebar, and agent discovery without a restart.
- Non-owners do not see Delete, and the backend refuses non-owner delete requests.
- Adds backend tests for manual/server delete, dependency listing, and the owner check, plus frontend tests for the dialog and post-delete navigation.

## Test plan
- `NODE_ENV=test pnpm test` green across the monorepo (core-backend, core-frontend, mcp-core, hexis-mcp), including new tests.
- `pnpm typecheck` clean across all 7 projects.
- Verified live: booted the app from source via `deployment/docker-compose.build.yml` against a scratch KB branch and exercised deleting a `.tool` manual and an MCP server entry as the owner, confirmed non-owner/non-reader/unauthenticated refusals, and confirmed the tool disappears from `/api/tools` and `/api/agent/all-tools` immediately with no restart.

---
## Summary by cubic
Owners can now delete a tool from its tool page; previously only entire plugins could be deleted. The owner-only flow shows dependencies, removes the `.tool` file or MCP server entry, wipes stored credentials, and refreshes discovery immediately; dependent skills remain with dangling `allowed-tools` references.

**Safeguards**
- Confirmation requires typing the exact tool name and shows readable dependencies plus credential counts, never values or identities.
- Incomplete dependency scans, unreadable plugins, non-owner requests, and externally managed plugins fail closed.
- Namespace matching handles doubled underscores and leaves ambiguous keys untouched, preventing similarly named tools from sharing a wipe.
- Vault cleanup retries and re-scans for concurrent writes, while reporting if deletion succeeds but cleanup fails.
- Manual deletes restore files only when no commit landed; a committed deletion that fails to push leaves the parked dotfile.
- Own-property checks prevent names such as `toString` from matching undeclared MCP servers.

<sup>Written for commit 6fa8e143f29ebedac0795f18bb33120c6af9039f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Bevel-Software/Hexis/pull/247?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


h3xxit:
All 14 cubic findings answered — 13 fixed in `4a5ffb48`, one disagreed with below. The branch is also synced with `dev` (`45935f9e`).

### Fixed

**`toString` and friends** (`tool-delete.service.ts` 234/283/299) — `name in map` is now `Object.prototype.hasOwnProperty.call`, in all three places a tool NAME indexes a map: the server-exists check, the `plugin.json` extensions half, and the dependent-plugin scan. Two tests cover it — a 404 instead of a delete against a file the server is not in, and an empty dependent list instead of every plugin that happens to declare servers.

**Restoring after a landed commit** (`:259`) — you're right, and the manual path was inconsistent with `deleteServer`'s own documented contract. The catch now restores only when nothing was committed, probed via `hasUnpushedCommits` (a probe that itself fails counts as "a commit may exist"). A landed-but-unpushed commit leaves the file parked, dot-prefixed and invisible to the scanner, for the pipeline's recovery to own. Tested both ways.

**Partial discovery presented as authoritative** (`:229`) — `dependents` no longer swallows a failed scan, and now also refuses while `discovery.unreadable` is non-empty. "No other plugin carries it" is the sentence the owner deletes on, so a scan that did not finish refuses (503) rather than reassures; the dialog already declines to offer Delete on a dependents read it could not make.

**Dependent plugins the caller cannot read** (`:237`) — each candidate is now gated by `canRead` on its folder before it is named.

**Credential cleanup after the definition commit** (`:154`) — partially. The wipe is retried (3 attempts), and its final failure now says the tool was deleted but its credentials were not, naming the action left, instead of a bare 500. I did **not** build a durable retry queue: there is no job store in this module's reach, and adding one is a larger change than this ticket. Flagging it as the residual gap rather than claiming it closed.

**Namespace deletion race** (`db-secrets-vault.service.ts:255`) — the membership rule cannot be expressed in SQL, so the ids must be read before they are deleted. Rather than serialize namespace writes (which would mean a lock across every vault writer), `removeNamespace` now re-scans and deletes until a pass comes back empty, bounded at 5, and reports what the delete's own `returning` says went — not what the scan saw. A row landing in the window is taken by the next pass. Tested with a store that gains a row between the scan and the delete.

**`null` mutation sentinel** (`:260`) — emitted only when a SHARED row actually went; otherwise only the affected user ids.

**Leading-underscore namespace collision** (`secrets-vault.contract.ts:125`) — real, but it is not the membership rule that creates it: `utcpNamespacePrefix` implements the UTCP SDK's own encoding (sanitize, then double every `_`), so `foo` + `_bar_KEY` and `foo_bar` + `KEY` are literally the same vault row before anything of ours runs. Rejecting leading-underscore variable names would break `.tool` files already storing under them, and we cannot change the encoding without desynchronising from the SDK. So the fix is at the destructive edge, where the consequence actually lands: `ToolDeleteService.claimedVars` drops any declared variable whose namespaced key another existing tool's prefix could also produce. The row survives, still reachable by the tool that still declares it, and the delete cannot take a live credential from a tool nobody asked to delete. Counting and wiping use the same claim, so the dialog's number matches what goes.

**Confirmation gate** (`DeleteToolDialog.tsx:84`) — added: Delete stays disabled until the tool's name is typed exactly (the name as the backend spells it, once it answers). The AC allowed either the typed name or the plugin dialog's bare confirm; the typed name is the right one here, because this dialog also wipes credentials that are not the clicker's.

**Exhaustive-sounding skills copy** (`:110`) — heading is now "Skills you can see that allow it" and the empty state "No skill you can see names it in its allowed tools", which is what the backend actually sends.

**Delete on an externally managed plugin** (`ToolPage.tsx:176`) — hidden. The item now needs `isOwner && linksAreManaged`, so it is not offered where the route can only answer 422.

**Test name vs. branch covered** (`tool-delete.service.test.ts:198`) — split in two: one for the unknown slug, one where `listAccessible` genuinely omits the tool, asserting the same 404 from both `deleteTool` and `dependents`.

### Disagreed

**Committing the whole plugin directory** (`tool-delete.service.ts:317`) — not changed. `McpServerEditService.putServer`, the existing writer of these same two files, commits `${kbDirName}/${PLUGINS_DIR}/${folder}` for exactly this reason: `mcp.json` and `plugin.json` must land together or not at all, and `CommitDriver.runPendingCommit` takes ONE path (`git add -- <path>`) with no explicit-path batch form. Splitting the delete into two commits would make the half-deleted state reachable — the extensions block re-declaring auth for a server that no longer exists, which the surrounding comment calls out. Narrowing this is a change to the commit pipeline's shape and to `McpServerEditService` with it, not to this endpoint.

Full suite green on the new head: `pnpm typecheck` clean across 7 projects, `NODE_ENV=test pnpm test` exit 0 (7098 tests).

h3xxit:
Second cubic run — all 5 findings fixed in `442ec988`. Branch re-synced with `dev` (`26965e0e`; one import conflict in `ToolPage.tsx`, dev's `NameWithBadges` beside this branch's `PageActions`/`DeleteToolDialog`, both kept).

Four of the five converge on one root cause, so they are fixed together.

### The ambiguous namespace key (`tool-delete.service.ts:255`, and with it `secrets-vault.contract.ts:125`)

You're right, and it invalidates the approach I took last round rather than just its edge case. `claimedVars` asked the tool catalog whether a colliding tool existed — but `listAllSummaries()` is `scan().manuals`, which omits every manual the scan could not read or could not parse. Its silence was never proof, so "no other tool claims this key" was a conclusion the data could not support, and the wipe went ahead on it.

So the ambiguity is now settled once, in the predicate, with no catalog involved: **`isKeyInNamespace` answers NO for every `_`-leading remainder**, with no `declared` exception. `foo` declaring `_bar_KEY` and a tool named `foo_bar` declaring `KEY` name the same vault row under the UTCP encoding, and nothing about the key says which — so for an operation whose only job is counting and WIPING a namespace, the ambiguous row is left standing. Recoverable, where one wrongly deleted is not. `countNamespace` and `removeNamespace` lose the parameter with it, and `ToolDeleteService` loses `claimedVars` and its `listAllSummaries` dependency entirely (net −80 lines).

The row is left out of the COUNT as well as the wipe, so the dialog's "N stored keys … are wiped" stays true to what actually goes.

This also closes the first run's P1 on `secrets-vault.contract.ts:125` properly rather than routing around it — there is no exception left to exploit.

### Stale `kind` in the tally (`db-secrets-vault.service.ts:271`)

Fixed as suggested: `DELETE … RETURNING` now returns `userId` and `kind`, and both the tally and the notification read those rows. The scan's copy is used for nothing but the id list. (`key` isn't returned — nothing downstream of the delete reads it; `tally` and the notify split both key off `userId`/`kind`.)

### Lost notification on an interrupted run (`:264`)

Fixed: notification moved into a `finally` over the whole loop, via a `notifyNamespaceGone` helper. A pass that throws does not un-delete the passes before it, and a listener that never heard would keep serving a pooled connection for a credential that no longer exists. Tested — the scan throws on pass 2, the run rejects, and `u-1` is still told about the row that went on pass 1.

### Dialog state surviving a navigation (`DeleteToolDialog.tsx:42`)

Fixed, and it was real: `deleteOpen` lives on `ToolPage`, which stays mounted across a `:slug` change, so the dialog saw the new slug with the old tool's dependents and typed confirmation still in state — armed. It now clears `dependents`, `loadError`, `error` and `typed` during the render that first sees a new `slug` (the render-phase reset, not an effect, so there is no window at all). `busy` deliberately survives: it belongs to a request already in flight.

The new test rerenders the dialog with a second slug while the second dependents read is still pending, and asserts Delete goes back to disabled, the old dependents are gone, and only the NEW name re-arms it. Verified to fail with the reset removed.

### Test mock conflating two contracts (`tool-delete.service.test.ts:129`)

Fixed — `listAllSummaries` returns the full fixture list, as in production. With `claimedVars` gone the service no longer calls it at all, so the mock is now purely there to keep `authorize`'s exists-but-unreadable test honest, which was your point.

---

`pnpm typecheck` clean across 7 projects; `NODE_ENV=test pnpm test` exit 0 on the new head — 7548 tests (core-frontend 282 files/4534 +1 skipped, core-backend 195/2689, hexis-mcp 19/214, mcp-core 7/111).

PR #250, opened 2026-09-17
Share-From-Cards — Every plugin row and skill card on the skills and tools index offers Share from its own menu

h3xxit:
Implements `Data/Engineering/Knowledge/Hexis/Tickets/Share-From-Cards.md`.

## Summary
- Every skill card on Everything, Owned by me, a group's page, and your own space now shows a "…" menu (hover/focus, or right-click) with **Open** and **Share**; Share opens Manage access on the skill's own folder, matching the skill page's Share.
- Every plugin row offers the same menu with **Open** and **Share**, targeting the plugin's primary folder. For a plugin the caller cannot read, the menu offers **Subscribe** instead, matching the locked page's request-access behavior.
- The "…" button is keyboard reachable (tab order after the card, Enter opens, arrows move, Escape closes and returns focus). Tool cards are unchanged.

## Testing
- `NODE_ENV=test pnpm install && pnpm typecheck && pnpm test` — all green (typecheck across 7 projects; mcp-core, hexis-mcp, core-frontend, core-backend suites pass).
- New/updated tests: `packages/core-frontend/src/modules/library/__tests__/card-row-menu.test.tsx`, `packages/core-frontend/src/modules/library/__tests__/PersonalPluginPage.test.tsx`, plus 3 new cases in `PluginPage.test.tsx`.
- Manual verification against a local Docker boot at `5476838d` covering all acceptance criteria (card/row menus, Share targeting the correct folder per surface, locked-plugin Subscribe flow, keyboard navigation, manager Remove control non-overlap, tool cards unaffected).

---
## Summary by cubic
Adds a "…" menu to every skill card and plugin row on the library index, so Share and request-access are offered on the item itself rather than only on the page you have to open first.

- Skill cards offer Open and Share; Share opens Manage access on the skill's own folder, always on the default branch the catalog was read from, and offers the walk up to a folder the skill inherits rules from instead of leaving granted access read-only there.
- Plugin rows offer Open and Share targeting the plugin's primary folder; for plugins the caller cannot read, the menu offers Subscribe instead, wired to the same request, toast, and Requested state as the locked plugin page.
- The button appears on hover and focus, is a tab stop after the card or row, opens at the pointer on right-click, and is fully keyboard-drivable (Enter opens, arrows and Home/End walk the list, and Escape, Tab, or picking an item closes and returns focus to the button).
- Tool cards are unchanged and carry no menu — access to a tool is decided at its plugin — and the card props now enforce that in the type.
- On manager cards, the corner "…" and the Remove overlay no longer overlap.

<sup>Written for commit 9d75e3e016c2cb6596d4f2eb36018ce4c6607b2b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Bevel-Software/Hexis/pull/250?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


h3xxit:
All eleven review findings answered on `a30859bf`. Nine are fixed in code; two I am pushing back on, with reasons. (The tooling here has no per-thread reply, so they are collected in one comment against the lines they were raised on.)

### Fixed

**`PersonalPluginPage.tsx:140` — the dialog wrote to whatever branch was open.** Correct, and it was true of `PluginPage`'s dialog too, which the skill cards now share. Both pages pin `workspaceId={encodeURIComponent(DEFAULT_BRANCH)}`, as `LibraryPage` and `SkillPage` already did. New tests on both render with a non-default ambient workspace and assert the dialog is handed the default branch, so the pin cannot be dropped silently.

**`PersonalPluginPage.tsx:142` and `PluginPage.tsx:377` — no `onManageAncestor`, and a comment that claimed otherwise.** Same root cause, fixed the way the comment described rather than by correcting the comment: both pages now hold the dialog's own `FileTreeEntry` (the shape `LibraryPage` and `SkillPage` hold) instead of a repo-relative folder string, keyed on `relativePath`, with `onManageAncestor` wired to the setter. `PluginPage` keeps a `setManageFolder(folder)` helper so its other four openers — the title row's Share, the join-request banner, the locked view — are unchanged, and they gain the walk too. A skill on a plugin page nearly always inherits its rules from the plugin folder, so this was the case the card's Share was most often opened for. Covered by "a card's Share can walk up to the folder the skill inherits from" and its twin on your own space.

**`ItemActionsMenu.tsx:239` — focus stranded after `Subscribe`.** Right: an action that opens nothing left focus on a removed node. Picking now closes AND hands focus back to the `…`, exactly as Escape does; an action that opens a dialog mounts it after this and takes focus for itself, which is the behaviour we want.

**`ItemActionsMenu.tsx:215` — Tab escaped an open menu.** Tab now closes it. Deliberately without `preventDefault`: focus moves to the trigger inside the handler and the browser reads the focused element afterwards, so its own Tab continues one step past the `…` rather than from the top of the document. Both paths are covered by a test that fails against the old handler.

**`card-row-menu.test.tsx:195` — Enter never exercised.** Was a synthesised click. Now `userEvent.keyboard('{Enter}')` on the focused trigger, with the arrow and Escape assertions unchanged after it.

**`card-row-menu.test.tsx:156` — pointer placement unasserted.** Now asserts the fixed wrapper's `left`/`top` are the pointer's `120px`/`240px`.

**`card-row-menu.test.tsx:257` — the folder-less plugin was not covered.** Both halves now are: no `onShare` (the page cannot address folders yet) and `summary.folders: []` (the index named none).

**Changeset `patch` → `minor`.** Agreed — this adds a capability.

### Disagreed

**`PluginRows.tsx:116` — gate the Share branch on `summary.canRead`.** This would make the row disagree with the page it opens. `PluginPage` renders the LOCKED view only when `summary && !summary.canRead && pluginItems.length === 0`, and `pluginItems` is `data.items.filter(i => isInPlugin(i, plugin))` — the same predicate `pluginEntriesFor` uses for `hasItems` in `member = summary.canRead || hasItems`. So a caller with partial item access gets the MEMBER page, whose title row offers Share on the primary folder; with the suggested gate their row would offer Subscribe to a plugin whose page shares. The ticket's criterion is that the row's Share does what the plugin page's Share does, and un-gated is what makes that true. Share is also safe for a non-reader by construction: the dialog resolves its own `canWrite` and renders read-only, which is the honest answer to "who is this shared with?".

**`PluginRows.tsx:118` — keep Open when there is no summary.** Deliberate, and the same rule the row with no identity follows (`entry.name === null` → no menu). The second verb is the menu's reason to exist; `Open` alone is the row's own click said in words, so a one-item menu adds a control and no capability. Nothing is lost while the answer is outstanding: `ItemMenuFrame` renders its wrapper unconditionally and only its contents are conditional, precisely so the trigger can appear once the index settles without remounting the row under someone's finger. The rule is stated in `actionsOf`'s docstring and now has a test for both ways of having no folder.

`NODE_ENV=test pnpm typecheck && pnpm test` green for the frontend: 161 files, 2154 tests (2150 before), plus `pnpm ds:check` with no regressions.

h3xxit:
Second review pass (run `e26fc080`) raised two findings on `a30859bf`. One is fixed on `9d75e3e0`; the other I am answering.

### Fixed

**`card-row-menu.test.tsx:233` — the pick half proved nothing, because it picked Share.** Correct, and the sharper half of this is that the test was only green because `onShare` was a `vi.fn()`: it documented focus-return-to-`…` as Share's contract, which is not Share's contract.

The pick now happens on a plugin row's **Subscribe** — the verb whose real action genuinely opens nothing, which is the case the handback exists for and the case the first review raised it about. Only `requestPluginAccess` is stubbed, so the verb runs its own handler, and focus is asserted a second time after the request settles and the row re-renders into `Requested`. The card keeps the Tab half, which is its own. Both docstrings now say what is true: a verb that opens something (Share's dialog) or navigates (Open) hands focus on from there, and that is the destination's to own, not the menu's. Reverting `closeAndReturn` to a bare `onClose` fails the new test.

### Disagreed

**`ItemActionsMenu.tsx:217` — focus after `Open` navigates.** Real, but not this menu's to fix, and not a regression:

1. **It is exactly what clicking the card does.** `Open` is the card's or row's own click said in words — the menu item and the card body call the same `onOpen`/`openEntry`. Clicking the card leaves focus on the card, which the route change unmounts. If `Open` from the menu landed focus on a destination heading, the same navigation would end in two different places depending on whether you clicked the card or picked the verb, which is worse than either.
2. **The old code ended in the same place.** Before this change, picking `Open` left focus on the menu item, which unmounted with the menu — `document.body`. After it, focus moves to the trigger, which unmounts with the route — `document.body`. No behaviour changed for `Open`; what changed is `Subscribe`, which leaves the card rendered, and there the handback is the whole point.
3. **Where `Open` does NOT navigate, the handback already works.** A proposed skill's `onOpen` opens its change-request dialog in place (`openItem` → `setReviewing`), and the card stays mounted — the trigger fallback is right for that path.
4. **The app has no route-arrival focus convention to hook into.** `useFocusHandoff` is deliberately scoped to a state swap inside one component ("two views share one column"), keyed on a state value the same component owns; it cannot reach a heading in a page that has not mounted yet. `SkillPage`'s and `KbPageHeader`'s `tabIndex={-1}` headings are handback targets for in-page swaps, not arrival targets. Every navigation in the app — cards, rows, breadcrumbs, the nav — has this property today.

So the honest fix is an app-wide route-change focus target, which is a separate piece of work: it belongs to the router or the page shell, applies to every link in the product, and would be the wrong thing to invent inside a card's menu for one of the dozens of ways to reach a skill page. Happy to raise it as its own ticket if you want it.

`NODE_ENV=test pnpm typecheck && pnpm test` green for the frontend: 161 files, 2155 tests.

PR #252, opened 2026-09-18
hx-admin-changes-own-password — Every account changes its own password from the Account page, and the deployment admin is told theirs is set in the environment

h3xxit:
Ticket: hx-admin-changes-own-password — Every account changes its own password from the Account page, and the deployment admin is told theirs is set in the environment

## Summary
- Reproduced first (see ticket log): an Admin-role account and a business user already change their password identically — no role-based bug. The real defect is the deployment admin (`ADMIN_EMAIL`/`ADMIN_PASSWORD`): with no stored hash, `changePassword` took the "SSO-only first password" path, accepting any current password and planting a second credential that didn't replace the environment one — later attempts using the real env password as "current" were then refused.
- `auth.service.ts`: a single `isEnvAdminEmail(email)` helper now backs `loginWithPassword`, `changePassword`, `listAccounts` and a new `toClientUser`. `changePassword` refuses the deployment admin outright (before the password policy check), so no session holder can plant a persistent hash for that account anymore.
- `AuthUser` gains an optional `isEnvAdmin` flag, carried through `/api/auth/me` and login responses (never the credential itself).
- `AccountPage.tsx`: when `isEnvAdmin` is true, the password form is replaced by a banner explaining the password is set via `ADMIN_PASSWORD` in the deployment environment and cannot be changed here — no fields, no save button. Admin-role accounts and business users still get the normal form.
- Tests cover all three account kinds on both backend and frontend (parameterised so the Admin-role and business-user paths can't drift apart again).

## Verification
- `pnpm typecheck` clean across all 5 packages; `NODE_ENV=test pnpm test` — 233 files, 5442 passed, 1 skipped, exit 0.
- Local Testing (separate pipeline step) additionally verified against a real local boot from source: all three account kinds exercised through the live `/api/auth/change-password` endpoint and the real `AccountPage` render from actual `/api/auth/me` payloads.

---
## Summary by cubic
Closes a security hole where any session holder could plant a second, lasting password for the deployment admin (`ADMIN_EMAIL`/`ADMIN_PASSWORD`) without knowing the password. The service now refuses to store a hash for that account — from the Account page's change and from an admin's "Set password" — and refuses to let a hash already on that row sign in while `ADMIN_PASSWORD` is set, making the environment password its only credential. Wrong-password attempts cost the same lookup and scrypt work for every address, so response timing cannot reveal which email is `ADMIN_EMAIL`; both surfaces explain the password is set via `ADMIN_PASSWORD` instead of offering the form or button.

- Admin-role accounts and business users always took identical paths; role was never the cause of the original report.
- An `isEnvAdmin` flag on `AuthUser` is derived from configuration on each read and never stored.
- Unsetting `ADMIN_PASSWORD` makes the account ordinary again — form, stored password, and login all behave as before.

<sup>Written for commit de82aa11b73e0e497b75283a63a379413ae9e9ba. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Bevel-Software/Hexis/pull/252?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


h3xxit:
All three review findings are addressed in `a93af12d`.

**P1 — `createAccount` still stores a hash for `ADMIN_EMAIL` (auth.service.ts:179).** Valid, and the more serious half of the bug: `POST /api/admin/accounts` takes an arbitrary email, so any *other* admin could aim "Set password" at the deployment admin and plant exactly the second credential `changePassword` refuses to create — one that signs in beside the environment password and survives rotating `ADMIN_PASSWORD`. `createAccount` now refuses that target, before the policy check so the refusal names the real reason, through the same `isEnvAdminEmail` definition; the message is now one shared constant, so the two surfaces cannot drift into explaining the rule differently. The User Accounts page drops the button on that row and says the password is set in the deployment environment rather than offering an action that can only fail — erasure, the row's other action, is untouched. Tests: refused for a differently-cased `ADMIN_EMAIL` with nothing written (no insert values, no conflict clause), refused before the policy check, and accepted again once `ADMIN_PASSWORD` is unset; on the page, no Set password on that row, delete still offered, the ordinary account beside it unaffected.

**P3 — `ROW` fixture omits `onboardingDone`.** Valid. A real row always carries the boolean (`NOT NULL` with a default), so every payload these tests read had it `undefined` and a regression dropping it from `toAuthUser` would have passed. `onboardingDone: false` added to `ROW`, and the `carries the flag to the client` test now pins the client payload's key set the way the `listAccounts` test pins its own, plus the value itself.

**P3 — changeset summary contradicts the exception.** Valid. The summary line now reads "Every account except the deployment admin…"; kept rather than the suggested "Accounts can change their own passwords", because the exception is the point of the change and the qualified sentence states it outright. The body also now records the `createAccount`/User Accounts half.

Verification on this head: `pnpm typecheck` clean across all five packages and both apps; `core-backend/src/modules/auth` 83 tests and the full `core-frontend` suite (161 files, 2171 tests) pass under `NODE_ENV=test`.

h3xxit:
Addressed in `79c10293` (on top of a sync merge with `dev`, `bcd73678`).

**P1 — a stored hash on the `ADMIN_EMAIL` row still signed in.** Accepted, and the finding is right about the consequence: refusing to *write* a hash only helps from the moment the rule exists, and a hash can already be on that row — planted before this change, or left behind by an ordinary account that only later became `ADMIN_EMAIL`. Rotating `ADMIN_PASSWORD` would have left it working, which is the exact second-credential problem the write refusal was for.

cubic offered two remedies; I took the first. `loginWithPassword` no longer treats the environment credential as a first attempt that falls through to the stored one — while `ADMIN_PASSWORD` is set, that email is answered from the environment and nowhere else, and a wrong password is refused without the row ever being read (the test asserts `db.select` is never called). The failing branch still runs one scrypt verification against the existing decoy hash, so this account costs what every other one does and response timing does not reveal which email is the deployment admin.

**On not clearing the hash.** The second option — deleting the stored hash — is destructive and irreversible, and it would contradict a property the rest of this change is built on: every env-admin fact is derived from configuration on each read, never stored, so unsetting `ADMIN_PASSWORD` hands the identity back and it becomes an ordinary account again. A deletion would make that handover lossy, and it puts credential destruction on the login path, which is a far worse thing to get wrong than a login that ignores a credential. Refusing gets the same guarantee across a rotation — while `ADMIN_PASSWORD` is set, at any value, only the environment password works — without that cost. Both halves are pinned by tests: the hash is refused while the variable is set, and accepted again once it is unset.

Verification on the pushed head: `pnpm typecheck` clean for all 7 projects; `NODE_ENV=test pnpm test` exit 0 — 422 files, 6310 passed, 1 skipped.

h3xxit:
Addressed in `de82aa11`.

**P2 — the env-admin branch skipped the `users` lookup, so wrong-password timing could disclose `ADMIN_EMAIL`.** Accepted. The previous commit's branch answered that one address without touching the table, leaving its failure one database round trip shorter than every other address's. The generic `Invalid credentials` only hides which emails exist if the *work* is generic too — that is what the decoy hash already buys for the scrypt half, and the new branch quietly gave up the database half. My comment on that branch claimed the cost was uniform, which it was not; the code now matches the claim rather than the other way round.

Fixed exactly as suggested: the lookup happens once, before the branch, for every address, and for the deployment admin its result is simply not consulted. The refusal was always about *which credential answers*, not about skipping work, so nothing else moves — the hash on that row stays as ignored as it was and the row is still never written on this path.

One behavioural note for the reviewer: a successful env-admin login now issues the lookup in addition to its existing upsert. That is a second query on one non-hot path, and the success path is not what needs to be indistinguishable — a successful login announces itself regardless.

Tests: the wrong-password path pins exactly one lookup for the admin address, an unknown address and an ordinary account alike; the stored-hash refusal now asserts the row *is* read while `captured.set`/`captured.values` stay empty.

Verification on the pushed head: build OK, `pnpm typecheck` clean for all 7 projects, `NODE_ENV=test pnpm test` exit 0 — 422 files, 6311 passed, 1 skipped.

PR #253, opened 2026-09-18
Download-Action-Follows-Permission — The file tree's Download action is disabled, with the reason, for a person without download permission

h3xxit:
Ticket: `Data/Engineering/Knowledge/Hexis/Tickets/Download-Action-Follows-Permission.md`

## Summary
- The tree's context menu now asks the server for an entry's access once when it opens (`GET /api/workspace/<ws>/access?path=...&kind=file|folder`), not once per row, and renders Download — or a folder's "Download as zip" — disabled with the tooltip "You don't have download permission for this item" when the verdict says no; otherwise enabled as today.
- Disabled is `aria-disabled`, not the native `disabled` attribute, so the tooltip stays reachable by hover and by keyboard focus; activation is refused in the click/keydown handler itself, which is what stops the keyboard path (menu key, Enter, Space) rather than only a mouse click.
- While the lookup is in flight, or if it fails, the item stays enabled — a flash of disabled on every menu open would be worse than the refusal this preflight exists to avoid.
- A 403 that still happens (permission changed since the menu opened) now surfaces as a dismissible inline notice under the row it's about, instead of a browser `alert()`.
- The backend gate is unchanged; this is a frontend-only preflight in front of an authority that already existed.

## Test plan
- `NODE_ENV=test pnpm test`: 236 files / 3710 passed, 1 skipped.
- `NODE_ENV=test pnpm typecheck`: all packages pass.
- `pnpm ds:check`: no regressions.
- `FileExplorer.test.tsx` (86 tests) covers a downloadable file, a read-only file, a folder, the keyboard path, and the late-403 path.
- Local Docker build from source, exercised with two real users (a reader without download vs. a user with download) against a live backend: menu-open fires exactly one access call per open; Download/"Download as zip" render `aria-disabled` with the tooltip for the denied caller and enabled with no extra request for the permitted one; keyboard activation (Enter/Space) on the disabled item fires zero download requests; revoking the grant while the menu was already open produced a real backend 403 with zero `window.alert` calls and the inline `role="alert"` notice instead; the backend `GET .../raw?download=1` gate itself is confirmed unchanged and still enforces 403/200 correctly.

---
## Summary by cubic
The file tree's Download action now follows the per-path download permission: the context menu asks the server once when it opens and disables the item with the reason when denied, instead of offering the action to everyone and surfacing a browser alert on the server's 403.

- Disabled items use `aria-disabled` (not `disabled`) so the tooltip stays reachable by hover and keyboard focus; the click and keydown handlers refuse activation.
- Items stay enabled while the lookup is in flight or fails, since the unchanged backend gate remains authoritative.
- The verdict is keyed to the entry it was asked about, so a stale answer from a previous selection never disables a Download the caller does have.
- Only the latest download attempt may write the inline notice, so overlapping downloads can't report each other's failures.
- A late 403 (permission changed after the menu opened) appears as a dismissible inline notice under the row instead of an `alert()`.

**Migration**
- No migration steps. The backend download gate is unchanged; this is a frontend-only preflight.

<sup>Written for commit 9d40504601a330bbc263f96be1589a7fa82cacc4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Bevel-Software/Hexis/pull/253?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


h3xxit:
All three cubic findings accepted and fixed in 9d405046. Branch also merged up to `origin/dev` (b41a0df1) so the PR builds against a current base.

**P2 — stale download outcome overwrites a newer one** (`FileExplorer.tsx`, `handleDownload`)

Real, and reachable: the menu closes on click but the ROW does not, so a second Download is one reopen away while the first is still in flight. A slow refusal landing after a fast success reported a failure the user had just watched succeed. Added a per-row sequence number; only the latest attempt may write `downloadError`. `console.error` still fires for a superseded throw — the log should not lie about what happened, only the notice is suppressed.

Covered by a new test, `ignores a superseded download outcome instead of reporting a stale failure`, which drives the out-of-order case explicitly (first click hangs, second succeeds, first then refuses). Verified it bites: with the guard removed it fails on the stale notice.

**P2 — `useDownloadVerdict` keeps a prior verdict across a key change** (`FileExplorer.tsx`)

Accepted. In the current wiring this is latent rather than live — rows are keyed by `relativePath`, so a path change remounts the node and takes the menu with it — but the hook shouldn't depend on its caller's keying to be correct, and the failure mode you describe is the bad one: a stale `false` disables a Download the caller *does* have, and stays disabled because the `.catch` deliberately leaves the verdict alone (default-allow on a transient lookup failure).

Fixed by storing the verdict *with* the question it answers (`workspaceId`/`path`/`kind`) and reading it back only while that question still stands. Chose keying over resetting-before-each-lookup because it costs no extra render, so there is no window in which the wrong state is painted at all.

**P3 — `aria-disabled` item still brightens on hover** (`Menu.tsx`)

Correct — the existing rule reset the background and left the foreground, so a disabled row lit up to `text-ink` as if it were usable. Added `aria-disabled:hover:text-ink-muted`, placed in the default-tone branch rather than the shared line: the hover it undoes is tone-local (the danger tone never changes text colour on hover), so the override belongs next to it. The attribute selector gives it higher specificity than the plain `hover:` rule, so it wins independently of stylesheet order.

Evidence: `FileExplorer.test.tsx` 87 tests pass (86 + the new one); `pnpm typecheck` green across all packages; `pnpm ds:check` shows no regressions (206 vs baseline 215).

PR #254, opened 2026-09-18
Download-Implies-Read — A download grant implies read, the way an edit grant does

h3xxit:
Ticket: `Data/Engineering/Knowledge/Hexis/Tickets/Download-Implies-Read.md`

## Summary

A `download` grant now confers `read`, the way an `edit` grant does. Previously, granting Download without Read was a dead combination: the raw-file route checked read before download, so the grantee could neither open nor save the file, and the Manage access dialog let anyone create that pairing.

Verbs now nest as: `owner` over `edit` and `download`, and `edit` and `download` each over `read`.

- The resolver folds `download` grants into `read` the same way it already folds `write`/`owner`. A `deny download` says nothing about read — matching the existing grant-only rule.
- Direct, group, role, plugin, and inherited download grants all confer read.
- The Manage access dialog shows Read as checked and greyed whenever Download is checked, on new grants and existing rows, with the verb summary reading "Can read, Can download". Choosing Download alone still writes a single `download:` line.
- The agent guide's access paragraph now states the nesting explicitly.
- Tests cover the resolver for download-only grants/denials, the route for read on a download-only path, and the dialog's folding behavior.

### Noted behavior change
PersonalSpacesStep no longer closes a personal space whose `access.md` says `download: everyone` — it now reads as a deliberate opening, following directly from the resolver table.

---
## Summary by cubic
A `download` grant now confers `read`, the way an `edit` grant does. A download-only grant used to be a dead combination: the raw-file route checks read before download, so the grantee could neither open nor save the file.

**Behavior changes**
- Verbs nest as `owner` over `edit` and `download`, and each of those over `read`; the fold is grant-only, so `deny download` never strips a separate read grant.
- The fold applies to every grant shape — direct, role, group, plugin token, inherited, or node frontmatter — and the plugin roster now shares the same nesting table, so a download-only plugin member also lands in the read principal.
- The Manage access dialog shows Read as checked and greyed under Download, writes a single `download:` line, and the Everyone clamp always has a read verb to send.
- `PersonalSpacesStep` now treats `download: everyone` as a deliberate opening; existing `download:`-only rules become live read grants on upgrade.

<sup>Written for commit c49bfed519cd287cd9e1f501e7f7f89817f38664. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Bevel-Software/Hexis/pull/254?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


h3xxit:
Review addressed in `de9938c1` (on top of a fresh merge of `origin/dev`). All four findings accepted — one was a real bug, three were wording/dead-code.

**P2 — [plugin principals kept a second copy of the fold](https://github.com/Bevel-Software/Hexis/pull/254#discussion_r4046177085)** — valid, and the only behavioural miss in the PR. `plugin-principals.ts` had its own private `sourceVerbsFor`, so a plugin whose `access.md` granted `download:` to someone left them out of the synthesised `plugin/<slug>/read` principal: the roster and the resolver disagreed about the same file. It now imports the shared table from `access-grammar.ts` and the local copy is gone, so the nesting is written down once. Covered by a new case in `access-plugin-principals.test.ts` — a download-only member of the plugin reads a skill granted to `plugin/GTM/read`, and still cannot write it.

**P2 — [the generated access.md note could be read as `deny download` denying read](https://github.com/Bevel-Software/Hexis/pull/254#discussion_r4046177092)** — valid. Took the point rather than the exact suggestion, so the note names the case a reader would actually wonder about:

> `` `deny X` `` takes away that verb ALONE: grants imply the lower verbs, denials never do — `deny download` says nothing about read, and leaves a separate read grant standing.

**P3 — [the Everyone download-only guard is now unreachable](https://github.com/Bevel-Software/Hexis/pull/254#discussion_r4046177106)** — valid, and removed rather than reworded. `grantVerbs` is non-empty exactly when `effectiveNewVerbs.read` is true now that all four boxes imply read, so the branch could not be entered. Worth saying why removal is the right half of your either/or: the clamp already silently drops the verbs `everyone` cannot hold (picking Can edit + Everyone has always granted plain read), and the dialog carries a standing notice under the chip — "Everyone makes this file publicly readable: it can only be granted read access" — so the user is told before the click, not after it. Download-only was simply the last pick that clamped to *nothing*; now it clamps like every other. New test locks it in: Everyone + Can download issues exactly one `read` grant and raises no failure banner.

**P3 — [the changeset says `edit` where the grammar says `write`](https://github.com/Bevel-Software/Hexis/pull/254#discussion_r4046177114)** — valid; a reader copying `edit:` into an access.md writes a key the parser ignores. Changed to `write` throughout, keeping the UI label in a parenthetical so the release note still connects to what the dialog shows: "the way a `write` grant does (`write` is the key an access.md carries; the dialog shows it as \"Can edit\")".

Full suite green on the merged head: 425 files, 6311 passed, 1 skipped; typecheck clean in all 7 projects.

h3xxit:
Second review pass addressed in `c49bfed5`. Base re-checked first: `origin/dev` is still `d4441ba8`, so the merge from the last round is current and the branch has no conflict with its base.

**P3 — [“standingly” in the new test's comment](https://github.com/Bevel-Software/Hexis/pull/254#discussion_r4046289109)** — valid, it isn't a word anyone writes. The sentence now reads "the dialog clamps it to read and says so plainly, up front, while the chip is up", which also makes the timing explicit — the notice is there before the click, which is the whole reason the after-the-fact failure was the wrong shape.

**On the still-unmarked [plugin-roster thread](https://github.com/Bevel-Software/Hexis/pull/254#discussion_r4046177085)** — that fix did land, in `de9938c1`: `plugin-principals.ts` no longer has a private `sourceVerbsFor`, it imports the shared table, and `access-plugin-principals.test.ts` covers a download-only plugin member resolving through `plugin/<slug>/read`. The thread carries no addressed marker because it anchors on `access-grammar.ts` line 74, and `de9938c1` touched `plugin-principals.ts` rather than the anchored file.

The durable answer belongs at that anchor regardless, so `c49bfed5` puts it there. `sourceVerbsFor` now says it is the only copy of the nesting and names all three callers that fold through it — `resolveAtPath`, the plugin roster synthesis, and `personal-spaces.step.ts` asking whether a space is already open:

> THIS TABLE IS THE ONLY COPY. … A second fold written out by hand somewhere else is how one of them comes to disagree with resolution about the same file — add a verb here, not there.

That is exactly the failure mode this review caught, so it is worth a reader finding it at the table rather than after the fact.

Full suite green on `c49bfed5`: 425 files, 6311 passed, 1 skipped; typecheck clean in all 7 projects.

PR #255, opened 2026-09-18
Key-Buttons-One-Size — Every action button on the tool page is the size Edit server and Save already were

h3xxit:
Ticket: `Data/Engineering/Knowledge/Hexis/Tickets/Key-Buttons-One-Size.md`

## Summary
All action buttons in the tool page's connection and server sections now use the small (`sm`) size, matching what `Edit server` and `Save` already used. No button variant (outline, quiet, primary) changed — only size. Rows keep one height with buttons aligned on the same baseline. The `tiny` chip links under "Powers these skills" (`ToolPage.tsx:268`) are outside the connection/server sections and were deliberately left unchanged.

## Changes
- `packages/core-frontend/src/modules/library/components/tool-page/ToolConnectionSection.tsx`
- `packages/core-frontend/src/modules/library/components/tool-page/ToolVarRow.tsx`
- `packages/core-frontend/src/modules/library/__tests__/ToolConnectionSection.test.tsx` — asserts size on each button in the connection section
- `.changeset/key-buttons-one-size.md`

## Verification
- `pnpm typecheck` — 0 errors across all 8 projects
- `pnpm test` — 422 files, 6302 passed, 1 skipped
- `pnpm ds:check` — no regressions
- Booted from source in a local Docker Compose stack, confirmed via the shipped bundle that every action button in the connection and server sections renders `size: sm`, and mutation-tested the new test to confirm it is non-vacuous

---
## Summary by cubic
Unifies the tool page's action buttons to the `sm` size, so connection and server rows no longer mix two control heights.

**Bug Fixes**
- Only the size changed; button variants and the `tiny` links under "Powers these skills" stay as they were.
- The connection section's test now asserts `sm` on every button and button-styled link across all row states, and names the actions each case expects so a branch that stops rendering a control fails the suite.
- The test derives size tokens from `buttonClasses` rather than hard-coding class strings, so it tracks future padding-scale changes.

<sup>Written for commit dd51e9f8ebc32203a6708b30dab25fc62b7d3dda. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Bevel-Software/Hexis/pull/255?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


h3xxit:
Addressed cubic's finding on `ToolConnectionSection.test.tsx` (the one-size suite) in dd51e9f8 — it was right, and the gap was exactly as described.

`notSmall()` reads nothing into a control that is not on screen, and every fixture carries Open Secrets, which is always small. So "nothing here is the wrong size" was also true of a section that had stopped rendering almost everything, and the `controls.length > 0` guard only ruled out the empty case.

Each case now names the actions it expects, by accessible name, before it says anything about their size:

- the settled tool: Test connection, Open Secrets, Reconnect, Replace client secret, Replace, Remove
- the tool waiting on keys: the banner's Set key plus the rows' Set key and Add key
- each sign-in branch, the oauth-manual banner, and the missing client secret
- the row editor: Save and Cancel *joining* a row that keeps Replace and Remove — which is the whole reason their height has to match, so the case now says that out loud

The `controls.length > 0` guard is gone, subsumed by callers that name every control they expect.

Proved non-vacuous by mutation: dropping the Remove button from `ToolVarRow` now fails two cases on their names, where before it changed nothing at all.

Branch also merged up to `origin/dev` (12cb1bda, clean). `pnpm typecheck` 0 errors, `pnpm test` 6481 passed / 1 skipped, 0 failures.

PR #257, opened 2026-09-18
Key-Status-Fresh-After-Save — Saving or removing a key reloads the library so cards and plugin banners show the new state at once

h3xxit:
Ticket: `Data/Engineering/Knowledge/Hexis/Tickets/Key-Status-Fresh-After-Save.md`

## Summary
Saving or removing a tool credential used to leave the rest of the Library stale: only the tool page re-probed itself, so a plugin's card, its "needs setup" banner and the sidebar count kept showing the pre-save state until a full browser reload. Every surface that can write a credential — the tool page, "Connect your tools", the Secrets vault, and the `.tool` editor panel — now announces a shared `bevel:tool-credentials-stale` event on landing (including after an OAuth sign-in returns, authorized or error), and the Library reloads both its catalog and its plugin summaries in response. A failed save announces nothing, so nothing reloads and nothing changes.

## Acceptance criteria
- [x] After a key is saved or removed on the tool page, "Connect your tools", or the vault, the library catalog and plugin summaries reload without a browser reload.
- [x] After an OAuth sign-in returns to the tool page (authorized or error), the same reload happens.
- [x] Going back to the plugin page or the tools list shows the tool as configured (or not) according to the new state.
- [x] A failed save changes nothing.
- [x] Tests cover save, remove, OAuth return, and back navigation.

## Testing
- `NODE_ENV=test pnpm test` — 6333 passed, 0 failed
- Typecheck clean across the monorepo
- Manual QA against a from-source boot: save/remove on tool page and vault, OAuth authorized/error returns, forced-500 failed save (no reload), back-navigation to plugin/tools list — all verified in a real browser with no page reload

---
## Summary by cubic
Saving or removing a tool credential now refreshes the rest of the Library at once — plugin cards, "needs setup" banners, and the sidebar count reflect the new state without a browser reload. Previously only the tool page re-probed itself, so those surfaces kept the pre-save state until a manual refresh. The `dev` merge is behavior-neutral; its only conflict was a test-file append.

**What changed**
- Every surface that writes a credential (tool page, "Connect your tools", Secrets vault, `.tool` editor panel) dispatches `bevel:tool-credentials-stale` on success and on OAuth returns — authorized or error.
- On "Connect your tools", a wipe that removes some keys before failing still announces — those keys are really gone — while a wipe that fails on its very first call announces nothing.
- The Library refetches its catalog and plugin summaries together, since the banner and sidebar counts derive from summaries.
- An event is used rather than a direct call because `/connect` and `/secrets` are shell routes without a mounted Library; an unheard announcement does nothing.
- Failed saves dispatch nothing, so no reload happens when nothing changed.
- Ships as a `patch` to `@bevel-software/platform-core-frontend`.

<sup>Written for commit 2868aa1b0722198168e2b21bc6b8c2acc5c36804. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Bevel-Software/Hexis/pull/257?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


h3xxit:
Both cubic findings accepted and fixed in `7d092941`. The branch is also merged up to `origin/dev` (`ec9204cd`) so CI can build the merge commit.

**P2 — partial wipe leaves the Library stale** (`ConnectToolsPage.tsx`, the `wipe()` in `onToggle`). Real. Skipping a tool that holds several per-user keys is one DELETE per key; if the second rejects after the first succeeded, that first key is gone for good but the announcement — which sat after the whole loop — never fired, so the catalog kept offering a key nobody has.

`wipe` now takes a `removed` callback and calls it after each DELETE that lands; `onToggle` announces (and refetches its own page) in a `finally` if anything at all was removed. A wipe that fails on its very first call still announces nothing, because nothing changed — that case has its own test too.

**P2 — `#error` OAuth return untested.** Right: the source has always announced on both fragments, but only `#authorized` was covered. Added `announces a failed OAuth return too, and consumes that fragment`.

One note on what the new tests assert. The error-banner text is deliberately not asserted in either new case, because `refresh()` ends with `setError(null)` on success: in jsdom the mocked `getConnectPending` resolves right after the error is set and wipes the banner. In the real app the round-trip is slower and the message renders — manual QA saw it. That ordering is pre-existing and untouched by this PR, so the tests assert the thing under review (one event, fragment consumed, both deletes attempted) rather than pinning a race.

Verified the halfway-wipe test is not vacuous: it fails against the pre-fix source and passes after. Full suite green — 6490 passed, 1 skipped; typecheck clean across the monorepo.

PR #259, opened 2026-09-18
hx-local-server-answers-first — The local MCP server answers its client at once, runs on current Node versions, and says in one sentence what is missing

h3xxit:
Ticket: `hx-local-server-answers-first` — The local MCP server answers its client at once, runs on current Node versions, and says in one sentence what is missing

## Summary
- The server now answers the client's `initialize` immediately and discovers the workspace's tools (manuals, plugin downloads, plugin servers, UTCP client) in the background, instead of doing all of that before `server.connect`. Only the two small reads the handshake result depends on (`GET /api/config`, agent instructions) still run in front of it. `tools/list`, `tools/call`, `prompts/list` and `prompts/get` wait for discovery and then answer as before.
- A discovery failure no longer kills the process: it's reported on stderr and `tools/list` returns a single `hexis_unavailable` tool carrying the reason (`tools/call` gets the standard MCP error). Anything the failed discovery had already built is released; only the MCP server survives.
- The supported Node engine range is widened from `>=22.13 <23` to `>=22.13 <23 || >=24 <25` — exactly the majors `isolated-vm@6.1.2` ships prebuilt binaries for (verified empirically: 22→abi127, 23→none, 24→abi137, 26→none). CI now runs the `hexis-mcp` suite on both 22 and 24, and a test fails if the constant, `engines`, the CI matrix and the docs ever disagree.
- On an unsupported Node major, or when the native module resolves but fails to load, the command prints one sentence naming the supported versions and how to select one, then exits non-zero — no stack trace.
- The Desktop agents connection hint (welcome page and External agent access) now tells GUI-launched clients (Cursor, Claude Desktop on macOS) that they may not see `npx` on PATH and to use the absolute path from `which npx`; the JSON snippet itself is unchanged.
- README and troubleshooting docs state the supported Node versions and the PATH note.

## Test plan
- `pnpm install && pnpm build && pnpm typecheck && pnpm test`: 4058 passed / 1 skipped across 252 files, exit 0.
- From-source `docker compose` boot of the real deployment, driving the compiled `dist/cli.js` over stdio: `initialize` answered in the hundreds of ms while discovery was held behind an HTTP proxy for 12s+; `tools/list`/`tools/call` correctly waited for discovery then returned real data.
- Discovery failure simulated via 503 from the tools endpoint: `tools/list` returned the `hexis_unavailable` notice, `tools/call` returned MCP error `-32603`, stderr carried the reason, process stayed alive.
- Preflight sentence verified on real Node 23.11.1 and 26.9.0 binaries (one line, zero stack frames, exit 1), and on Node 24 with a truncated `isolated-vm` prebuild (native-load failure sentence).
- Node 24 native sandbox verified directly: `loadNativeSandbox() === 'loaded'`, `preflight() === null`, full suite (16 files / 178 tests) green under the real Node 24 binary.

---
## Summary by cubic
The local MCP server now completes `initialize` before expensive tool discovery, instead of blocking `server.connect` on manuals and plugins. Discovery-dependent requests still wait for the catalog; failures keep the server connected and expose a `hexis_unavailable` tool with the reason.

- Supports Node 22.13+ and 24 (the majors `isolated-vm` ships prebuilt binaries for), with a preflight that prints one actionable sentence and exits non-zero for unsupported versions, prereleases of supported ones, native load failures, and `@utcp/code-mode` installs where the addon does not resolve; embedders that install without the extras still pass.
- Runs the `hexis-mcp` suite on both supported Node majors and fails if the engine range, CI matrix, and docs disagree.
- Updates Desktop agent guidance for GUI clients missing `npx` — pointing at a supported version's `npx`, since `node -y` reads `-y` as its own option — while the snippet stays unchanged.
- Bounds `shutdown()`'s wait on an uncancellable discovery phase, releases what it builds afterward, and answers a `tools/list` racing shutdown with a going-away notice instead of an empty catalog.
- Adds coverage for answer-first handshakes, discovery failures, prerelease refusal, teardown, and preflight behavior.

<sup>Written for commit fb04ed793345400faac7a824150062d90be86c0e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Bevel-Software/Hexis/pull/259?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


h3xxit:
Review addressed. All fourteen cubic findings, in the order they were raised — thirteen fixed in code, one fixed at a different line than it was anchored to.

**P1 — `initialize` promised to a fresh keyless run** (`packages/hexis-mcp/README.md`): correct, and the ordering is in `cli.ts`: `establishOAuthConfig` completes before `createHexisMcpServer`, so on a first keyless run there is no transport to answer on until sign-in finishes. The paragraph now opens with *“Once it is signed in…”* and names the exception (the credential caches, so every later start is immediate).

**CI-matrix regex could read another job's matrix** (`supported-node.test.ts`): scoped. Both assertions now go through one `hexisMcpJob()` helper that slices the workflow at `\n  hexis-mcp:` and fails if that job is absent.

**README says “Node 22” while `engines` excludes 22.0–22.12** (×2, plus the repo README): fixed at the source rather than in each doc. `SUPPORTED_NODE` now carries a floor per major (`22.13`, `24`); `enginesRange()` builds `>=22.13 <23 || >=24 <25` from it, `supportedNodePhrase()` builds `22.13+ or 24`, and the docs test asserts that phrase in all three documents. The `engines` field is unchanged — it was already right; what was wrong was everything that claimed to agree with it.

**Preflight accepted 22.0–22.12** (`preflight.ts:149`): it no longer does. The check is major-plus-floor, compared numerically (string order would put `22.5` after `22.13`), and `preflight.test.ts` pins 22.5.0 and 22.12.9 as refusals with 22.13.0 as a pass.

**`node` cannot replace `npx` with those arguments** (`README.md:13`): right — `node -y @bevel-software/hexis-mcp` reads `-y` as its own option. Every place that said “point `"command"` at that version's `node`” now says `npx`, absolute, with `args` unchanged, and says why `node` will not do. The preflight sentences say the same. They also now suggest `nvm install 22` rather than `nvm use 22`: with a floor inside a major, `nvm use 22` can land back on the very version that was just refused.

**An absolute `npx` still goes through `/usr/bin/env node`** (`agent-clients.ts:83`): accepted, with one deliberate difference from the suggestion. The absolute `npx` stays the first instruction — it is what the acceptance criteria ask for, it is the fix for the common case (Homebrew/nvm `npx` with a `node` beside it that the client's own resolution then finds), and the JSON snippet has to stay machine-independent. What is added is the second failure by name: if the client then reports `env: node: No such file or directory`, the same folder holds `node`, so add it to the client config's `"env"` PATH. Both the welcome-page hint and the External agent access note say it, as do the README and the troubleshooting guide. Rewriting the snippet to an absolute `node` plus `npx-cli.js` would hard-code two machine-specific paths where one already works for most readers.

**`which npx` is not a command on Windows** (`ExternalAgentAccessPage.tsx:326`): `where npx` is named, in that paragraph and in all three documents and the welcome-page hint.

**Unescaped `discoveryError` on stderr** (`server.ts:611`, and the duplicate anchored at `:364`): escaped through `printable()`. The `:364` anchor is the agent-instructions log, which interpolates nothing — it is static text — so the fix landed only at the one site that carries a reason. `printable` moved from `core-backend/src/shared` into `mcp-core`, which both MCP surfaces already depend on (core-backend re-exports it, so its ~20 call sites are untouched) — the local server could not import it where it was, and a second copy is a second escaper that drifts. The client-facing copies keep the reason verbatim: inside a JSON-RPC string it is data, not a log line. New test: a reason carrying `\n` and a CSI sequence comes out as one escaped stderr line and an unmangled `hexis_unavailable` description.

**Unbounded `await ready` in `shutdown()`** (`server.ts:649`): bounded. `DISCOVERY_SHUTDOWN_GRACE_MS` (5s, inside the CLI's own 15s watchdog so it is the thing that acts first) caps the wait; past it, teardown proceeds, says so on stderr, and attaches `ready.then(releaseDiscovered)` so whatever the uncancellable phase finishes building is released when it comes to rest — `releaseDiscovered` is idempotent and re-reads `client`, so the late call frees a late child rather than double-closing an early one. The race uses an `unref`'d timer, so the losing side cannot hold the process open. Tested against a discovery phase that never returns.

**Shutdown during discovery answered `tools/list` with an empty catalog** (`server.ts:546`): the list handler now checks `closed` after the gate and raises `ConnectionClosed`, matching the call handler, rather than showing the reader the tool-less-workspace look the notice exists to prevent. Tested: shutdown, then a listing held behind discovery, then release — the listing rejects with `ConnectionClosed`.

**Missing `isolated-vm` let the CLI proceed** (`preflight.ts:126`): `preflight({ requireNativeSandbox: true })` from `cli.ts`, whose next statement imports `server.js` and through it code-mode's module-level `isolated-vm` — an `ERR_MODULE_NOT_FOUND` stack trace one line away. It gets its own one-sentence refusal. The default stays lenient for embedding hosts that installed without the extras, which was the original reason for `unresolved`.

**Hard-coded `nvm use 22` in the suite** (`preflight.test.ts:48`): derived from `SUPPORTED_NODE_MAJORS[0]` like the implementation does.

Green locally: `pnpm build`, `pnpm typecheck`, and `pnpm test` — 252 files, 4058 passed / 1 skipped; the `hexis-mcp` package is 17 files / 184 tests. No new lint errors in the touched files.

h3xxit:
Second cubic pass (against `09d256b7`) addressed — both new findings fixed in code, plus the one carried over from the first review.

**P2 — a prerelease of a supported version is accepted** (`preflight.ts:89`): valid, and it slipped through exactly as described — `22.13.0-nightly…` compares equal to the released `22.13.0`, so the floor check waved it past. `isPrerelease()` now refuses any version carrying a prerelease tag, with the same sentence the wrong-major and below-floor cases get. Two reasons, and they agree: semver excludes prereleases from `>=22.13 <23` unless the range names one, so `engines` never promised it; and `isolated-vm` publishes against a settled ABI, which a nightly of a major is not. Tested on `22.13.0-nightly…`, `24.0.0-rc.1` and `22.14.0-pre`, with the releases they are prereleases of still passing.

**P3 — the shutdown/`tools/list` test cannot tell which mechanism rejected** (`discovery-teardown.test.ts:134`): correct, and worse than "cannot tell" — I printed the actual rejection and it was `MCP error -32000: Connection closed`, the transport's, not the handler's. Deleting the guard left the test green. Fixed by restructuring rather than by asserting harder: the listing is issued **first** and given time to park on the gate, so it is the handler that resumes when discovery lets go, and only then does shutdown start. The assertion now also names the message only the handler writes (`hexis-mcp is shutting down.`). Re-verified by mutation — with the `if (closed) throw` guard removed the test fails, with it restored it passes.

**P2 — unresolved `isolated-vm` lets the CLI proceed** (`preflight.ts:126`, carried over from the first review): the CLI stopped proceeding in `09d256b7` — `cli.ts` calls `preflight({ requireNativeSandbox: true })`, and an unresolved module returns a one-sentence refusal there. Since the finding kept its footing, I have closed the wider case too rather than only the CLI's: an unresolved sandbox is now a refusal for **any** caller whenever `@utcp/code-mode` itself is installed (`codeModeInstalled()`), because in that install the import that dies is one `server.js` makes on its own — an embedding host would hit the same `ERR_MODULE_NOT_FOUND`, and `requireNativeSandbox` is not something it knows to pass. What still passes is the install the leniency was written for: no code-mode, so nothing here ever reaches for a sandbox. A test asserts `codeModeInstalled()` is true in this repo, so that guard is live rather than theoretical.

Green locally on the new head: `pnpm build`, `pnpm typecheck`, `pnpm test` — 252 files, 4058 passed / 1 skipped; `hexis-mcp` is now 17 files / 187 tests.

PR #262, opened 2026-09-18
Pending-Tools-Shown — Show a tool proposed in an open change request as a Proposed card

h3xxit:
Ticket: `Data/Engineering/Knowledge/Hexis/Tickets/Pending-Tools-Shown.md`

## Summary
- Backend lists pending tools: open change requests, visible to the caller, that add a `.tool` manual or an MCP server entry, exposing the tool's name, target plugin, and change request number.
- Library renders each pending tool as a Proposed card (visually distinct from active tools) that links to its change request and cannot be opened as a tool.
- After a change request merges, the card is replaced by the real tool on next reload; after it closes/is cancelled, the card disappears.
- Callers without visibility into the change request see no card.

## Tests
- Backend listing covers visible, invisible, merged, and closed change requests.
- Frontend covers Proposed card rendering and linking.
- `NODE_ENV=test pnpm test`: typecheck 7/7 projects done; 254 files, 4080 passed / 1 skipped, exit 0; design-system ratchet shows no regressions (206 vs baseline 215).
- Local boot from source (docker compose build) against a scratch KB branch verified `/api/tools/pending`, card rendering with IN REVIEW badge, merge/reject transitions, and visibility scoping across roles.

---
## Summary by cubic
Tools proposed in open change requests now appear in the library as Proposed cards, so authors and eligible approvers can review them before merge. Previously, the tools UI showed nothing because the catalog only read the default branch; proposed tools remain review-only and are not callable or counted as integrations.

**Backend**
- Adds `GET /api/tools/pending` for new `.tool` manuals and `mcp.json` server entries under the plugins root.
- Limits results to open requests targeting the default branch and users who authored the request or can write the proposed path.
- Removes proposals after merge, closure, or cancellation, and matches released tools by UTCP namespace.
- Shares branch-reading and visibility logic with proposed skills, including escaped warnings for failed proposal reads.

**Frontend**
- Shows dashed `In review` cards that link to the change request, show both the flavor and review badges, and never open a tool page; `NameWithBadges` keeps long names readable.
- Applies normal plugin membership, linked-root, and personal-space rules without counting proposals as setup work.
- Renames `PendingSkillReview` to `PendingItemReview` for shared skill and tool review dialogs.

<sup>Written for commit 3f7b0510f0d9b6658adc0d7865c62e4888f0d273. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Bevel-Software/Hexis/pull/262?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


h3xxit:
Review addressed in `ea775dd4`. Twelve of the thirteen findings landed as code; one is answered below.

**Fixed in code**

- **`tool-manuals.contract.ts:401` (P2, slug not route-safe/unique)** — the pending slug is now the resolved manual name, set the same way `scanDisk` sets it (`descriptor.slug = descriptor.name`), so `weather.tool` declaring `id: get_weather` and a file named `My Weather.tool` both line up with the tool that replaces the card. Card uniqueness moved off the slug: the React key now carries the declaration path too.
- **`pending-tools.service.ts:62` (P2, `-` vs `_` duplicate)** — the released set and the proposal are both compared through `utcpNamespacePrefix`, matching the catalog's own dedup identity. Covered by a new test where a proposed `on-call` server is dropped against a released `on_call`.
- **`library-data.tsx:287` (P2, linked-root membership)** — `plugins: pluginsHoldingTool(t.path, pluginSummaries)`, and that value feeds `pluginOfItem` as it does for released tools.
- **`pending-proposals.ts:72` (P2, non-default base)** — `visibleProposedFiles` now keeps only open requests whose `base` is `DEFAULT_BRANCH`, for both surfaces. Everything downstream is stated in terms of that branch, so a request aimed elsewhere was a card that could never resolve.
- **`PersonalPluginPage.tsx:86` (P2, no-op click)** — you were right that the comment and the card contradicted each other. The page now opens the review dialog, identically to the gallery and the plugin page; a card that names the reader as the reviewer and then does nothing was the one arrangement that could not be right.
- **`pending-proposals.ts:129` (P3, `readAt` conflation)** — `readFileAtRef` already answers `null` for genuine absence, so everything reaching that catch was infra. It still returns `null` (the module's contract is "never throws") but logs a warning first, so the shelf degrades observably.
- **`tools.api.ts:175` (P3, `null` body)** — `Array.isArray(body?.tools)`.
- **`tool-manuals.service.ts:798` (P3, `baseName` doc)** — the doc now says what it is (the parser's provisional fallback) rather than what it is not; the divergence it described is gone with the slug fix above.
- **`LibraryPage.tsx:250` (P3, stale comments)** — updated on all three pages that hold `reviewing`.
- **`.changeset` (P3, patch → minor)** — both packages now `minor`.
- **`pending-tools.service.test.ts:136` (P3, fail-closed untested)** — correct: the harness answered an explicit `false` per path, which is the denial branch, not the `.catch`. `canWriteBatch` now rejects for that test, with the ordinary "tree says nobody writes" denial split into its own case.
- **`tool-manuals.pending.route.test.ts:88` (P3, no spy)** — `expect(listPendingTools).not.toHaveBeenCalled()`.
- **`pending-tools.test.tsx:167` (P3, no real load)** — the test now re-renders the *mounted* tree against the new mock answer rather than mounting a fresh one, so a provider that cached proposals across a reload would fail it.

**Answered, not changed**

- **`library-data.tsx:278` (P3, Tools band count vs `toolCount`)** — I don't think there is a divergence to reconcile here. The Skills band already counts proposals the same way (`shownSkills` derives from `skillItems`, which includes pending ones), so the Tools band is mirroring the established behaviour rather than departing from it. The place the skills side *does* guard is the "N skills in this plugin are not shared with you" note, which is computed against `releasedSkillItems` precisely so a proposal cannot mask a hidden skill — and that guard is intact. There is no tool equivalent of that note, and `summary.toolCount` is never rendered beside `toolItems.length`: its only consumer on this page is `DeletePluginDialog`, where released-only is the right number, since a proposal has no file on the default branch to delete. The sidebar row and the band do differ for the life of a proposal, but they differ for skills too, and "a card is here that the plugin's total has not counted yet" is what `In review` on the card already says.

Tests: `pnpm typecheck` 7/7, `NODE_ENV=test pnpm test` 4258 passed / 1 skipped, exit 0. `pnpm ds:check` 209 vs baseline 215, no regressions. Branch synced with `origin/dev` first (`161492b2`); that merge needed `pendingTools: []` added to two `LibraryData` fixtures dev introduced.

h3xxit:
Both findings from the re-review of `ea775dd4` are fixed in `6ee50b71`, on top of a fresh `origin/dev` sync (`13f3994f`).

- **`pending-proposals.ts:151` (P2, unescaped log interpolation)** — valid, and I should have caught it when I added the warning: the path and the branch come off a change request somebody else opened and the reason comes off git, so all three are attacker-shaped. They go through the shared `printable` now, matching every other log site in this backend. `printable` quotes as well as escapes, so the hand-written quotes around the path are gone rather than doubled.

  New suite `shared/__tests__/pending-proposals.readAt.test.ts` pins it on a capturing logger: silence when the file is simply absent (`readFileAtRef` already answers `null` for that, so only infra failures reach the catch), a warning when the read actually fails, and no raw newline or escape byte reaching the sink. Verified it fails against the unescaped version.

- **`pending-tools.test.tsx:323` (P3, spy not restored on failure)** — also valid. The spy comes off in `afterEach` now. Scoped to that one spy rather than `vi.restoreAllMocks()`, which I tried first and which tears down the `vi.fn()` module mocks `beforeEach` depends on (6 of 8 tests in the file went red).

Verification on the new head: `pnpm typecheck` 7/7 Done; `NODE_ENV=test pnpm test` 269 files, 4312 passed / 1 skipped, exit 0; `pnpm ds:check` 209 vs baseline 215, no regressions; `pnpm lint` 170 problems, exactly the pre-existing count (my first draft of the escaping test added a `no-control-regex` error, rewritten as `includes` checks).

The earlier `library-data.tsx:278` Tools-band-count finding still stands as answered rather than changed — reasoning in my previous comment.

(Correcting a miscount there, for the record: it opened "twelve of the thirteen findings" where the round was thirteen of fourteen. The itemised list under it was complete.)

h3xxit:
The one finding from the re-review of `aa1277b2` is fixed in `18d046ee`, on top of a fresh `origin/dev` sync (`1d141b04`).

- **`LibraryCard.test.tsx:213` (P3, stale doc comment)** — correct, and the comment was left behind by my own previous commit: gating the flavour badge behind `!pending` is exactly what made "it still says how it is declared" untrue. The comment now says the flavour badge is deliberately absent and why — a proposal has no file to edit yet, and the change request the card opens shows the reviewer the declaration itself.

  I also gave the test something to check. It was asserting nothing either way about the flavour badge, so the claim lived only in prose; it now asserts `MCP server` is absent on the `mcp` flavour, with the `utcp` side already covered by the badge-count test below it.

Verification on the new head: `pnpm typecheck` 7/7 Done; `NODE_ENV=test pnpm test` 279 files, 4495 passed / 1 skipped, exit 0; `pnpm ds:check` 210 vs baseline 215, no regressions; `pnpm lint` 171 problems — identical with and without my change (verified by stashing), so the one that appeared since my last pass came in with `origin/dev`, not from here.

That clears every thread on this PR. For the record across the three rounds: 17 findings, 16 fixed in code, 1 (`library-data.tsx:278`, Tools band count vs `summary.toolCount`) answered with a reasoned disagreement earlier in this thread rather than changed.

PR #263, opened 2026-09-18
hx-non-md-creation-without-read — Creating a non-markdown file where you cannot read warns you first, and a proposed row opens its request at that file

h3xxit:
Ticket: `hx-non-md-creation-without-read` (Data/Engineering/Knowledge/Hexis/Tickets/Non-Md-Creation-Without-Read.md)

## Summary

- Creating or uploading a non-markdown file into a location the user cannot read now shows a dialog first — "You won't be able to see <name> after it is added: you don't have read access to <folder>." with Continue/Cancel; Cancel creates nothing. Markdown files (which carry the creator `read:` grant in frontmatter) are unaffected. One gate covers the sidebar new-file box, drag-and-drop, and the upload menu, asking once per multi-file batch and listing the affected names.
- Clicking a proposed row now opens the change-request dialog at the file that was clicked (selected in the file tree, diff shown), with the request and file encoded in the URL (`?cr=<n>&file=<path>`) so a reload lands on the same file.
- Investigation reproduced two of the three reported symptoms against a running instance and both are fixed here:
  - **Concurrent proposals 409ing**: two proposals racing the same "do I have an open request" read both tried to open one. The second now joins the first in flight instead of erroring, and a genuine server-side duplicate refusal is adopted (the existing request is used) rather than surfaced as a failure.
  - **Slow change-request list loading**: listing open requests previously fetched refs once per open request. It now refreshes the clone once per list read and diffs locally, so cost no longer scales with the number of open requests. Freshness guards ensure a `fresh=1` read still forces a refresh, and a branch unknown to the clone still fetches for itself.
  - The third symptom ("request not visible to its submitter") did not reproduce with read permission present; nothing was changed for it.

See `.changeset/non-md-creation-without-read.md` for full detail.

---
## Summary by cubic
Fixes ticket `hx-non-md-creation-without-read`: creating a non-markdown file you won't be able to read now warns first, proposed rows open at the clicked file, and two reproduced change-request bugs (concurrent proposals 409ing, slow list loading) are fixed.

**New Features**
- One Continue/Cancel dialog gates the new-file box, drag-and-drop, and upload menu — asked once per batch, listing affected names, with Cancel creating nothing; the markdown exemption is exact-case, matching the backend grant.
- The gate always answers its caller, even across a workspace switch or unmount, and a second question supersedes the first instead of stranding it.
- The dialog lands on the clicked file from `?cr=<n>&file=<path>`, and a stale `file` link falls back to the request's first file.

**Bug Fixes**
- If the joined proposal flow fails, the second proposal retries on its own; a server-side duplicate refusal is adopted as the existing request rather than shown as an error.
- The list refreshes the clone once per read and diffs locally instead of fetching per open request; fresh reads still force a refresh, and branches unknown to the clone still fetch for themselves.
- A dropped folder colliding with an existing unreadable folder gets no grant and no warning — a documented limitation, since deciding requires reading the very folder at issue.

<sup>Written for commit db56deaf4bf1a3f69a61687e51a1039f90e532b2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Bevel-Software/Hexis/pull/263?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


h3xxit:
Addressed cubic's ten findings on `8b30622`. Nine are fixed in code (pushed as `db56dea`); one is answered below.

### Fixed

**`isMarkdownName` is case-sensitive** (`unreadableCreate.ts:31`, P2) — correct and important: `CreatorAccessService.planForCreate` splices the per-file grant for `rel.endsWith('.md')` only, so `NOTES.MD` is folder-governed, gets no grant, and vanishes. It is now `/\.md$/`, and the test that asserted `isMarkdownName('NOTES.MD') === true` now asserts the opposite, with `unreadableAmong(['a.md', 'A.MD'])` alongside it.

**The gate always answers its caller** (`unreadable-create.context.ts:84 / 90 / 92`, P1 + P2 + P3) — all three are the same hazard (a promise an upload is waiting on, left pending), and all three are closed:
- the question is stamped with the workspace it was asked in, exactly as the move/delete confirmation above it already is; a branch or workspace switch resolves it `false` and closes the dialog;
- a read check still in flight when the tree unmounts, or when the workspace changes under it, answers `false` instead of opening a dialog nobody can see;
- a second gate-passing action supersedes the open question rather than orphaning it — the superseded batch resolves `false`, so it creates nothing.

New `state/__tests__/unreadable-create.context.test.ts` drives each of those directly, plus the three short-circuits that must keep answering `true`.

**A joined proposal gets its own attempt when the one it joined fails** (`propose.api.ts:110`, P2) — applied to both in-flight maps. The map entry is already cleared by the time the rejection surfaces (the `finally` is attached first), so the fallback is a fresh attempt, not a re-join of a dead promise. The dropped-connection-after-create case now does what the finding describes: the second flow's own open meets the duplicate refusal and adopts the request. Two tests.

**`changedPathsForPr` resolves the refs once** (`git.service.ts:2533`, P3) — with `fetch: false` the skip decision now resolves both branches to COMMITS and the diff runs on those commits, the same pinning `changedFilesForPr`'s `at` uses. `knowsBothBranches` became `publishedPrCommits`; `resolvePublishedBranchRef` keeps its contract and now shares one `rev-parse` with it, so the pinning costs nothing. A real-git test drives `update-ref -d refs/remotes/origin/<head>` inside the mutex window — it fails with `unknown branch` without the change and passes with it.

**A composed concurrency test** (`propose.concurrent.test.ts:94`, P3) — fair: the maps were only tested apart. Two `proposeKnowledgeChange` calls now race through the whole flow before the slow list read answers either, asserting one list read, one branch create, one open and two writes. A key mismatch between the two maps fails it.

**`.git` in the `ensureRemotesFetched` fixture** (P3) — added, and worth saying what it uncovered. With the accidental lazy-clone bootstrap out of the way, `force still joins a fetch already in flight` failed for a real reason: on a COLD workspace-directory lookup the first fetch settles and is cleared inside the microtasks the second caller spends resolving its directory, so there is nothing to join. The join is sound once the lookup is warm, which it is for every request after a process's first, so the test now warms the lookup and holds the fetch until both callers have looked. Exactly the coupling the finding predicted.

**The stale `?cr=&file=` seed** (`ChangeRequestDialog.tsx:214`, P3) — the seed holds while the detail is in flight (that is the landing it exists for) and the detail then decides: a pick the request does not contain is dropped for the ordinary first-file landing, so an old link opens the request instead of a pane reading a path the request no longer has. The early read for an unconfirmed seed is kept deliberately — it runs concurrently with the detail fetch and is what makes the legitimate landing immediate; in the stale case it is one wasted read whose result is never shown.

### Answered, not changed

**A dropped folder colliding with an existing unreadable folder** (`unreadableCreate.ts:52`, P1). The finding is right about the backend: `planForCreate` seeds on the topmost segment that is not on disk, so dropping `Reports/` into an existing `Data/Reports` produces no new directory, no grant, and no warning.

It is not answerable from the client, and both approximations are worse than the gap:

- Deciding whether the directory exists means asking what is inside a folder the caller has no read on. That is the one question the access model declines to answer — `GET /access?path=&kind=folder` computes from the access config and cannot see the disk — and an existence oracle there would leak the contents of unreadable folders.
- Warning about every nested name instead would warn about files that WILL be visible: in the ordinary folder drop the directory is new and the grant is seeded. That is the same broken promise pointed the other way, and it would fire far more often than the case it covers.

So the gap is recorded where the next reader will hit it (the module docstring now names it and says why the client cannot see it) rather than papered over. In that case the behaviour is the pre-ticket behaviour: the file is created, it reaches the folder's owners, and the server logs the ungranted create. Closing it properly means the server answering "would these be visible to you after creation?" for a batch — a preflight mirroring `planForCreate` — which is a card of its own, not a change to slip into this one.

Full run at the new head: build, `pnpm typecheck` and `pnpm test` all clean — 4 projects, 6750 tests, 0 failures.

PR #265, opened 2026-09-18
Plugin-Display-Name-One-Rule — Every way of creating a plugin derives and stores its display name by the same rule

h3xxit:
Ticket: `Data/Engineering/Knowledge/Hexis/Tickets/Plugin-Display-Name-One-Rule.md`

## Summary

A plugin's name now comes from one place. Its identifier is the manifest's `name`; what people see it called is the manifest's `displayName`, and — when the file says nothing — that same `name`. The folder it lives in is no longer consulted for either.

Previously the manifest carried `displayName` only when it disagreed with the folder's spelling, the rename path deleted the field the moment it agreed again, and every reader fell back to the folder — so the API always answered with a display name while the file sometimes omitted it, and where a plugin happened to sit was a hidden input to what everyone saw.

Every write path now persists the field:
- Creation (the New plugin dialog and the `create_plugin` tool, one endpoint) stores the name its creator typed, trimmed, whether or not it matches the identifier or the folder; both answer with `name` and `displayName` exactly as written.
- A display-name edit stores what was typed, never removes the field, and never touches the identifier.
- A new startup step (`PluginDisplayNamesStep`) backfills each existing plugin's folder spelling into its manifest only where the manifest lacks the field and the folder is spelled differently from the identifier — so nothing renames itself on upgrade. A manifest that already has the field, or whose folder already matches its identifier, is left alone.

## Test plan
- `pnpm --filter core-backend vitest run` — 254 files / 4080 passed, 1 skipped, 0 failed
- Targeted suite over the 9 touched files (plugin-creation-display-name, plugin-names, plugin-rename.service, kb-plugin-source, steps, plugins.routes, plugin-provision.service, plugin-index.service, compile-marketplace): 204 passed / 0 failed
- Typecheck clean across all 7 workspace projects; `pnpm build` clean
- Local from-source docker compose boot (`deployment/docker-compose.build.yml`): verified backfill (first boot reports changes, second boot reports none), creation via REST and via the `create_plugin` MCP tool, PATCH rename behavior, and that GET /api/plugins matches every manifest exactly regardless of folder name — all 6 acceptance criteria confirmed live, then torn down with `down -v`

---
## Summary by cubic
Makes a plugin's display name follow one rule everywhere: the manifest's `displayName`, falling back to its `name`. The folder is no longer an input; previously the field was written only when it differed from the folder's spelling, and every reader fell back to the folder, so where a plugin lived silently changed what it was called.

**What changed**
- Creation and rename always persist a trimmed `displayName`; creation returns it, and a blanked rename falls back to the identifier.
- `pluginDisplayNameOf` no longer takes a folder argument, and `renderPluginManifest` always writes the field.
- Bundle discovery keeps its folder rule but resolves it into the manifest it synthesizes, so discovery and the shared reader agree.
- `ensurePersonalPlugin` reports the existing manifest's display name instead of the folder spelling.

**Migration**
- New startup step `PluginDisplayNamesStep` backfills the folder spelling into manifests that lack a usable `displayName` and whose folder differs from what the manifest already shows.
- It runs after `PluginManifestsStep`, is idempotent, and skips bundles, symlinked manifests, and authored names.

<sup>Written for commit 5f79f06626df341829a23e88bfec17d0aaaaa5e0. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Bevel-Software/Hexis/pull/265?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


h3xxit:
Review addressed. Thirteen of the fifteen findings are fixed in code (commit pending push); two are answered here.

**Fixed**

- `PluginDisplayNamesStep` — a present-but-**blank** `displayName` is now treated as absent. The new reader falls through a blank field to `name` where the old reader fell through to the folder, so a blank field in a differently-spelled folder was precisely the manifest that renames itself on upgrade (`P2`, kb-layout.ts:416 thread).
- `PluginDisplayNamesStep` — the skip test compared the folder against the *identity*; it now compares it against what the manifest **alone shows** (`displayName`, else `name`). That covers the plain case (`gtm/` + `{"name":"gtm"}`) and the one the identity missed (`My Plugin/` + `{"name":"My Plugin"}`, identity `my-plugin`), which used to be rewritten to say what it already said.
- `PluginDisplayNamesStep` — a `plugin.json` that is a **symlink** is no longer read or written: `lstat` + `isFile()`, the same judgement discovery makes. A bundle folder with a linked manifest can no longer queue a write the runner refuses.
- `withDisplayName` copies with `Object.defineProperty`, so an own `__proto__` key from `JSON.parse` is copied as data instead of becoming the object's prototype.
- `ensurePersonalPlugin` reads the existing manifest and answers with its `displayName`, so the idempotent second call cannot report a name the file disagrees with.
- `bundle.source.ts` carries the **same** display name into `manifest.interface` that it writes into `manifest.displayName`, so the compiled Codex `interface` and the catalog cannot disagree.
- `PROVISIONED_OUTPUT.displayName`'s description now holds for `my_plugin` as well as `create_plugin`.
- `renderPluginManifest`'s second parameter is **dropped** — on the creation path the folder's leaf *is* the typed name, so no production caller ever passed two values, and a second parameter is only a way for them to disagree.
- Fixtures in `compile-marketplace`, `plugin-index.service` and `plugins.routes` tests now spell `displayName` differently from the folder (`Go To Market`, `Marketing Platform` in folder `Mktg`, `Google Tag Manager` in folder `GTM`), so a regression to the folder fallback fails them.
- The stray U+0014 in the provision test comment, and the rename test's comment that misdescribed the backfill's own rule.
- New coverage: blank-field repair, the non-identifier-name skip, the symlinked manifest, `__proto__`, the padded bundle `interface`, and an ensure answering from an edited manifest.

**Answered, not changed — the two trim findings** (`pluginDisplayNameOf` returning `declared.trim()`, `renderPluginManifest` trimming what it writes):

The acceptance criterion this PR implements is that creation persists the typed name **trimmed**, and the rename path obeys the same rule. Every write path therefore stores a trimmed value, which makes the trim on read a no-op for anything this platform has written — the "exactly as persisted" guarantee on creation and rename responses holds either way. The only case the two differ is a hand-edited manifest with padding, and there trimming is what I want: `"  Sales  "` and `"Sales"` are the same name, and answering with the padding would push it into every client's picker. Leading and trailing spaces are not part of what a plugin is called, so no reader here treats them as meaningful.

Full backend suite green: 256 files, 4094 passed, 1 skipped, 0 failed. Typecheck clean across all 7 projects.

PR #266, opened 2026-09-18
Principal-Chip-Truncates — Long name in Manage access field truncates inside its chip

h3xxit:
Fixes ticket `Data/Engineering/Knowledge/Hexis/Tickets/Principal-Chip-Truncates.md`: a long name in the Manage access field truncated inside its chip instead of the chip staying bounded and truncating with an ellipsis.

## Summary
- Bound each principal chip to the field's width so the label truncates with an ellipsis and the remove button always stays visible at the end.
- The full name remains available via a tooltip and as the remove button's accessible label.
- Verified with long unbroken strings and at 200% browser zoom that the chip stays intact (no overflow of the field).

## Test plan
- [x] `pnpm install && pnpm build && pnpm typecheck && pnpm test` — green (252 backend test files, 170 core-frontend test files; 4061 passed, 1 skipped)
- [x] New visual regression test `ManageAccessDialog.chip-truncation.test.tsx` covering a long email and a long group name
- [x] Booted the app from source in a container and measured real Chromium layout at 100% and 200% zoom against the built stylesheet — 0 violations across 30 cases plus a 160–900px width sweep; confirmed the pre-fix markup reproduces the original overflow bug as a control

---
## Summary by cubic
Fixes the Principal-Chip-Truncates issue: long names in the Manage access field no longer overflow their chip or the field. Each chip is bounded by the field width, its label truncates with an ellipsis, the remove button stays visible at the end, and long chips wrap onto new lines. The full name remains available as the chip’s tooltip and the remove button’s accessible label.

- Adds a patch changeset for `@bevel-software/platform-core-frontend`.
- Tests compile the app’s own stylesheet and assert the chip’s CSS declarations, so a purged utility, a missing Tailwind layer, or a broken CSS build fails instead of staying green.

<sup>Written for commit 855fbdb465304b5b1465db2840d35678b7f5dfb5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Bevel-Software/Hexis/pull/266?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


h3xxit (commented):
Answering the three cubic findings. Two are fixed in `aeb52f14`; the third I believe is wrong, and I've said why rather than acting on it.

---

**1. `ManageAccessDialog.chip-truncation.test.tsx` — "asserts only that utility classes are present; a stylesheet-level regression leaves all three green" — accepted, fixed.**

The point stands: `toHaveClass('truncate')` is not a layout assertion, and it pinned the contract to exact utility strings, so restyling the chip would have failed the tests for no behavioural reason.

The suite now asserts the declarations the element's own classes actually generate. `__tests__/tailwindDeclarations.ts` compiles this package's `src/index.css` with this package's Tailwind and maps each class on the element to its rule, so the assertions read `max-width: 100%`, `min-width: 0px`, `overflow: hidden` / `text-overflow: ellipsis` / `white-space: nowrap`, `flex-shrink: 0`, `flex-wrap: wrap` — the CSS the browser is handed, not the strings in the JSX. One compile for the suite, ~100ms.

Checked by mutation rather than asserted:

| change | result |
| --- | --- |
| drop `truncate` from the label | 3 tests fail on the missing `text-overflow: ellipsis` |
| drop `max-w-full` from the chip | 3 tests fail (`expected undefined to be '100%'`) |
| drop `shrink-0` from the remove button | 3 tests fail (`expected undefined to be '0'`) |
| swap `truncate` → `overflow-hidden text-ellipsis whitespace-nowrap` | still passes, as it should |
| point `index.css` at a missing `@import` | 3 tests fail on the broken stylesheet |

So a renamed or withdrawn utility, a Tailwind upgrade whose output changes, or a CSS build that no longer compiles now fails the suite. To be exact about the limit: the candidates are fed in from the rendered element, so this does not simulate a content-scan purge (these classes are static literals in the JSX, so that risk is nil here), and it is not a renderer — it says what CSS the box is given, not what the box then measures. Pixels stay the visual check's job; the real-Chromium sweep in the test plan remains the authority for the rendered result.

**2. `.changeset/principal-chip-stays-in-the-field.md` — "Nothing in the chip is sized in pixels" is inaccurate — accepted, fixed.**

Correct, and it contradicted the sentence before it: the remove icon is a deliberate `<X size={12} />`. The note now reads "What bounds the chip and truncates its label is sized in percentages and rem rather than pixels", which is the claim the fix actually supports.

**3. Third test reuses a captured `input` that React remounts when the chip list grows — disagreed, no change.**

The premise doesn't hold, on two counts.

The input is not remounted. `pickedChips.map(...)` is a single array child, which React reconciles as its own slot; the `<input>` that follows it keeps its position among the parent's children whether the array holds 0 chips or 2, so it is never matched against a chip. Measured directly against this component: after the first chip is added, `document.querySelector('input')` is the *same node* as the captured reference (`SAME_NODE= true`, `STILL_CONNECTED= true`), and the suite's third test passes — the second `user.type` does reach React's `query` state, the second suggest fetch does fire, and the second chip is found.

The suggested remedy would break the test. The placeholder is conditional — `placeholder={pickedChips.length ? '' : 'Add people, groups, roles or plugins…'}` (`ManageAccessDialog.tsx:1425`) — so once a chip exists, re-querying by placeholder text finds nothing and the test fails where it currently passes. I confirmed that too: `findByPlaceholderText` after the first pick throws `Unable to find an element with the placeholder text of: …`. Holding the reference is the correct thing to do here precisely because the node survives while its placeholder does not.

---

Green after the change: `core-frontend` 174 files / 2381 tests, package typecheck clean, `eslint` clean on both test files, `pnpm ds:check` 206 vs baseline 215, no regressions. Branch also merged up to `origin/dev` (`f639966b`) before the fix commit.

h3xxit (commented):
Second round: both findings on `aeb52f14` are accepted and fixed in `75c27de2`.

---

**`tailwindDeclarations.ts:75` — "silently returns `{}` when the utilities layer is missing" — accepted, fixed.**

Right, and the failure mode was the bad one: the helper exists to keep the assertions honest, so degrading to empty results made the parser's own breakage look like a chip regression (and would have passed an absence assertion vacuously). `layerBody` now throws, naming the stylesheet and saying whose fault it is:

> No `@layer utilities` in the compiled stylesheet: …/src/index.css no longer emits the layer this parser reads, so no element's declarations can be resolved. This is a fault in the test helper (most likely a Tailwind upgrade), not in the component.

On the variant half, I took the substance but not the literal remedy, because merging a varianted rule into an element's declarations would be wrong — `hover:text-danger` styles the button *on hover*; reporting it as a declaration the button has would make the helper lie in the other direction. What was genuinely wrong is that "styled, but only under a variant" was indistinguishable from "Tailwind emitted nothing for this class". Those are now separated: `declarationsOf` collects every class Tailwind emitted *any* selector for (including `.hover\:text-danger:hover`), and throws on a class that produced no CSS at all —

> Tailwind emitted no CSS for `truncat` on \<span>. The utility is purged, renamed or misspelled — the element is not styled the way its class list says it is.

— which incidentally gives the suite the purged/renamed-utility check that the first round could not claim. A variant class is now known to exist and contributes nothing unconditional by design; if a future chip relied on `sm:truncate`, the positive assertion still fails, naming the missing property.

**`ManageAccessDialog.chip-truncation.test.tsx:124` — "`0px` exemption is only valid for `min-width`" — accepted, fixed.**

Correct: `width: 0px` and `max-width: 0px` are real pins that would collapse the chip, and the filter dropped them. The exemption is now `min-width` only, and the filter carries the property through so a failure names it.

**Mutation-checked, not asserted:**

| change | result |
| --- | --- |
| parser reads a layer that is not emitted | 3 tests fail with the helper's own message |
| `truncate` → `truncat` on the label | 3 tests fail with "Tailwind emitted no CSS for `truncat`" |
| `max-w-0` on the label | 3 tests fail (`expected [ [ 'max-width', '0px' ] ] to deeply equal []`) — passed before this commit |

The first round's mutations still hold: dropping `max-w-full`, `truncate` or `shrink-0` each fails all three tests, and swapping `truncate` for `overflow-hidden text-ellipsis whitespace-nowrap` still passes.

**Still open from round one:** the "stale `input` reference" finding, which I answered rather than acted on — the node survives the chip list growing (same node, still connected; React reconciles `pickedChips.map(...)` as its own child slot), and the suggested re-query would break the test because `placeholder` is cleared once a chip exists (`ManageAccessDialog.tsx:1425`). Detail in the [previous reply](https://github.com/Bevel-Software/Hexis/pull/266#pullrequestreview-5253786721).

Green on `75c27de2`: `core-frontend` 174 files / 2381 tests, package typecheck clean, eslint clean on both files, `pnpm ds:check` 206 vs baseline 215. Branch is up to date with `origin/dev`.

h3xxit (commented):
Third round: the one finding on `75c27de2` is accepted and fixed in `855fbdb4`.

**`tailwindDeclarations.ts:126` — `styledClassNames` scans raw CSS, so declaration values register as selectors — accepted, fixed.**

The finding is right, and the doc comment was the giveaway: it claimed "every class Tailwind emitted a selector for" while the code held "every name that occurs after a dot somewhere".

I measured it on this stylesheet with the chip's own class list rather than take it on faith. The old scan admits **41** names; a selector-only scan admits **27**. The 14 extras are exactly what the finding predicted — values and theme variables (`3.3`, `25rem;`, `5px;`, `05);`, `16);` …) — and one of them is class-shaped: **`com`**, out of the `tailwindcss.com` banner comment.

So the collision is constructible, not theoretical. Putting the genuinely absent utility `com` on the chip's label:

| helper | result |
| --- | --- |
| pushed (`75c27de2`) | **3/3 pass** — the absent utility slips through, exactly as the finding says |
| this commit | 3/3 fail: `Tailwind emitted no CSS for `com` on <span>` |

One correction to the finding's reasoning, for the record: "every candidate is a member of the set whenever Tailwind emits anything dotted" overstates it — membership is by exact string, and most junk tokens carry trailing punctuation (`25rem;`), so they cannot collide with a real class name. It took a token like `com` to make it reachable. The conclusion stands regardless; the mechanism just had a narrower mouth than described.

The fix is the one suggested: the stylesheet is walked **per rule**, brace-matched, and text is read as a selector only where it precedes a `{` and is not an at-rule prelude — a `;` or `}` ends whatever was being read, so a declaration value can never register a class as styled. The utilities layer and its bare-class rules now come out of that same walk, which also retires the separate string-index scan. And the missing-class check now runs *before* the layer check, so an element whose classes are all unknown is reported as unknown classes rather than as a missing utilities layer.

Every guard still bites (each fails all three tests):

| mutation | result |
| --- | --- |
| utilities layer not recognised | helper's own message, naming the stylesheet |
| `truncate` → `truncat` | `Tailwind emitted no CSS for `truncat`` |
| absent utility `com` on the label | `Tailwind emitted no CSS for `com`` (new) |
| `max-w-0` on the label | `expected [ [ 'max-width', '0px' ] ] to deeply equal []` |
| drop `max-w-full` / `shrink-0` | `expected undefined to be '100%'` / `'0'` |
| `truncate` → `overflow-hidden text-ellipsis whitespace-nowrap` | still **passes**, as it should |

Green on `855fbdb4`: `core-frontend` 174 files / 2381 tests, package typecheck clean, eslint clean, `pnpm ds:check` 206 vs baseline 215. Branch up to date with `origin/dev`. The shipped component and stylesheet are untouched by this round — `git diff a44b710d 855fbdb4 -- ManageAccessDialog.tsx src/index.css` is empty.

Still open from round one: the "stale `input` reference" finding, answered [here](https://github.com/Bevel-Software/Hexis/pull/266#pullrequestreview-5253786721) rather than acted on.

PR #267, opened 2026-09-19
Rename-Never-Overwrites — A rename or move onto an existing name is refused instead of overwriting the file

h3xxit:
Implements Data/Engineering/Knowledge/Hexis/Tickets/Rename-Never-Overwrites.md.

## Summary
- Renaming or moving a file or folder onto a path where a file or folder already exists is now refused with "A file named <name> already exists in <folder>." (or "A folder named…"), leaving the existing entry untouched.
- The sidebar rename box shows this message inline and stays open; a refused drag leaves both entries in place and shows the same message.
- The agent's move and copy tools refuse with the same sentence.
- Case-only renames on a case-insensitive filesystem are still allowed (not treated as a clash with the entry itself).

## Tests
- File onto file, file onto folder, folder onto file
- Drag clash
- Agent move and copy clash
- Case-only rename allowed
- Ordinary rename unchanged

pnpm typecheck clean; core-backend vitest 255 files/4087 passed; core-frontend vitest 173 files/2383 passed (run with NODE_ENV=test).

---
## Summary by cubic
Implements Rename-Never-Overwrites: renaming, moving, or copying onto a taken name is refused instead of replacing the file or merging into the folder. The refusal names what is in the way — “A file named Notes.md already exists in Sales.” or “A folder named …” — and is a 409 from `move_file` and `copy_file`.

**Behavior**
- The rename box stays open with the sentence under the row; a refused drop leaves both entries in place.
- The move and copy land atomically, so a destination claimed after the check is refused too; without hard links, an exclusive create claims the name, and a lost race no longer deletes the winner's file.
- `move_file` and `copy_file` check write access before the destination, so callers who cannot write a folder cannot learn whether a name in it is taken.
- Case-only renames are still allowed when the folder lists both spellings as one entry, including a re-cased parent folder; a move onto a hard link under another name is refused.
- A refused move creates no empty destination folders.
- `copy_file` previously replaced an existing destination file; it now refuses, and the agent replaces content with `write_file`.
- Replacing a file still means uploading a new version by the same name.

<sup>Written for commit 4a230406092e0e2bba02f8f24a2c3c7bfdf9d6dd. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Bevel-Software/Hexis/pull/267?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


h3xxit:
Every finding is accepted and fixed in code; nothing here is a disagreement.

**The three P1s — the look and the landing were not the same moment.** Right: a destination created after `assertDestinationFree` / `existingAt` returned was still replaced, on all three paths. Locking could not close it (the sidebar's `moveEntry` deliberately takes no path lock — holding both ends is the one shape that deadlocks here), so the guarantee now comes from the filesystem call instead, in one place: `packages/core-backend/src/shared/rename-no-replace.ts`.

- A file moves as `link` + `unlink` — `link` is the atomic no-clobber primitive, and the source only goes away once both names exist.
- A folder claims its name with `mkdir` first (`EEXIST` if anything is there) and renames onto the empty folder it owns; a rename that fails anyway gives the claim back, so a refused move leaves no folder behind. Windows skips the claim — it refuses a taken directory destination natively.
- A copy lands with `COPYFILE_EXCL` (or the `wx` flag on the validated-bytes path), under the destination's lock.
- Filesystems with no hard links, and `EXDEV`, fall back to what was there before, so nothing that worked stops working.

Whichever call fires, the destination is re-read afterwards so the refusal carries the same sentence the look would have produced — a lost race and a plain clash are indistinguishable to the caller. The look stays: `EEXIST` alone cannot say *file* or *folder*.

**P2, the identity exemption was too broad.** Agreed — two hard links are one inode under two names a user can see separately. A case-only rename is now recognised by inode **and** by the two spellings folding together, in `inspectDestination` and in the tools' `existingAt`. New tests at both layers cover a move onto a hard link under an unrelated name.

**P2, editing while the rename is in flight.** Taken. The box is `readOnly` for that window and `onChange` returns early, so it keeps focus and the caret but the answer lands on the name it was actually about. New test freezes a request mid-flight and asserts one call, one name.

**P2 / P3, the two hard-link setups on a case-insensitive disk.** Both fixed. The tools test runs only on a case-sensitive disk (`!caseSensitiveDisk`, like its sibling); the service test tolerates the `EEXIST` a case-insensitive disk gives, because that `EEXIST` *is* the state it wants to test.

Branch synced with `dev` first. `pnpm typecheck` clean; core-backend 256 files / 4098 tests green, core-frontend `FileExplorer` 119 green.

h3xxit:
Second pass — all five findings on `214fe12a` accepted and fixed; no disagreements.

**P1, the fallback was a replacing rename** (`locking-filesystem.ts`). Correct, and it was the worst possible place for it: filesystems without hard links would have kept the original overwrite bug, unseen by any test. The fallback now claims the name with `open(…, 'wx')` — atomic and no-replace everywhere Node runs — and renames onto its own claim, removing the claim if the rename then fails. No path in the module reaches a replacing rename any more except the self-rename, where the source and the destination are the same entry.

**P2, lowercased path equality does not prove a case-only rename** (`workspace.tools.ts`). Agreed, and it was wrong in the direction that matters: on a case-sensitive disk `deal.md` and `Deal.md` fold together as strings while being two real entries. Identity plus folding was still two wrong readings stacked. The distinguisher is the directory — a filesystem that folds the two spellings has **one** entry to list, one that does not has two — so the parent is read and the two basenames counted. That is now the single `isSelfRename` used by the service, by the move and by the tools' `existingAt`, so preflight and landing cannot answer differently. The tools test that asserted the opposite (a hard link under a case-variant name being allowed) was wrong and now asserts the 409.

**P2, dangling link vs. `exists()`** (`locking-filesystem.ts`). Taken — a link to nothing owns the directory entry, fails the exclusive create, and reads as absent to a following probe. Switched to `lstat`. New test creates exactly that and asserts the clash sentence for both copy and move.

**P3, the folder test passed vacuously.** Right: `inspectDestination` refuses before `moveFolder` is ever entered, so neither the `mkdir` claim nor the `rmdir` rollback ran. `landFolder` and `claimThenRename` are exported now and driven directly, which is the only way to reach either — the claim-then-fail path asserts no folder is left behind, and the fallback asserts a taken name is refused rather than replaced. The old test is reworded to what it actually verifies.

**P3, the hard-link clash was Linux-only.** Agreed, the scenario holds on APFS and NTFS too. `skipIf` dropped. The neighbouring case-only rename test now probes the disk (`foldsCase()`) instead of guessing from `process.platform`, and asserts the right pre-state on either kind.

Branch synced with `dev` first (unchanged since the last merge). `tsc --noEmit` clean; core-backend 256 files / 4109 passed, 1 skipped.

h3xxit:
Third pass — both findings on `d6e31819` accepted; no disagreements.

**P2, a re-cased parent read as a different folder** (`rename-no-replace.ts`). Valid, and it is the same mistake one level up: I compared the parent paths as strings, but the volume that folds `notes.md` into `Notes.md` folds `Sales` into `sales` right alongside it, so `Sales/notes.md` → `sales/Notes.md` is one folder, one entry and one rename. The parents are now compared by filesystem identity (string equality first, since that is the answer almost every time and costs nothing; `lstat` dev/ino otherwise), and only then is the folder listed to count the two spellings. A genuine move between two different folders of one hard-linked file is still refused — asserted in the same test.

**P3, the changeset sentence.** Fair — "one that does not lists two" makes a reader parse a subject-verb disagreement before reaching the second clause, and this is published changelog for three packages. Split into two clauses that cannot collide: "a filesystem that folds the two spellings shows a single entry, while one that keeps them apart shows two."

One thing worth flagging for whoever runs this locally, unrelated to the review: syncing with `dev` brought the plugin-display-name work, and `packages/shared/dist` in a workspace built before that merge still declares the old two-argument `pluginDisplayNameOf`, so `tsc` reports four phantom errors in the plugin files until `packages/shared` is rebuilt. Nothing in the source; just a stale artifact.

Branch synced with `dev` (merged `b1489e40`, clean). `pnpm typecheck` clean across all packages; core-backend 258 files / 4131 passed, 1 skipped.

h3xxit:
Fourth pass — the one finding on `9c7f13b6` accepted; I only half-fixed the previous round.

**P2, a re-cased parent with the same filename is still refused.** Correct. Last round taught `sameFolder` to compare parents by identity, which got `Sales/notes.md` → `sales/Notes.md` right, but `Sales/notes.md` → `sales/notes.md` then fell through to the listing check — and with both basenames the string `notes.md`, `listed.has(old) && listed.has(new)` is the same question asked twice, so the single entry read as two and the rename became a clash with itself.

The listing exists for exactly one purpose: telling two spellings apart. When the basenames are identical there are no two spellings, so the pair is the source itself and the check is skipped — the same short-circuit `oldAbsolute === newAbsolute` already had, one level up. Both are now documented together as the two cases that never reach the listing because there is nothing in them to count.

One honest note on coverage: that branch is only reachable on a case-folding volume (on a case-sensitive disk the two parents are different directories, so `sameFolder` returns false before it), so the test asserts `self` and performs the rename where the disk folds, and asserts the destination is simply free where it does not. CI on Ubuntu takes the second half; macOS and Windows take the first.

Branch synced with `dev` (already up to date at `b1489e40`). `tsc --noEmit` clean; the shared, kb-fs and workspace suites 52 files / 1577 passed.

h3xxit:
Fifth pass. cubic reports 0 issues on `30898f89`, so the open request here is the architecture review's one violation — accepted, not disputed.

**`copy_file`/`move_file` reported destination existence before the write-access check.** Real, and the dry run made it cheap to reach. On a protected branch `move_file` computed `writeBlocked` only when there was *no* collision, and ranked "A file named rules.md already exists in Locked." above the write refusal — so a caller who may not write `Locked/` could ask for a name and read off whether it was taken. `copy_file` was worse: it had no preflight of its own at all, leaning on the lock gate, which refuses only *after* the copy has already answered.

That sentence is a fact about the destination folder, so it is now said only to someone who has earned it:

- `move_file` settles `writeBlocked` first and then looks at the destination at all — its on-disk spelling, its platform-name verdict and its occupancy are each computed behind that gate, and the refusal order (`managed` → write → existence) follows. Gating the spelling too matters: `"<name> is a platform file name"` would otherwise still differ by whether the name exists.
- `copy_file` checks the destination with the same `writeBlocked` before it looks, and raises the same `AccessDeniedError` the lock gate would.

The leak was in the tools' own preflight, not in the feature: the sidebar route acquires both path locks — the same access gate — before `moveEntry` looks at anything, so it never had this shape. Nothing changes for a caller who may write the destination.

Tests pair a taken destination with a free one in a folder the caller cannot write and assert the two are indistinguishable — same status, same words up to the path each names — for the real call and for the dry run.

Branch synced with `dev` (already up to date at `b1489e40`). `tsc --noEmit` clean; core-backend 258 files / 4145 passed, 1 skipped.

PR #268, opened 2026-09-19
hx-restrict-keeps-person-visible — Every access row has one menu: pick a verb set or Deny, and restrictions below the parent are written in the background

h3xxit:
Ticket: hx-restrict-keeps-person-visible (Data/Engineering/Knowledge/Hexis/Tickets/Restrict-Keeps-Person-Visible.md)

## Summary
Restricting someone below what a parent folder grants no longer makes them disappear from the Manage access dialog, and no longer takes a detour through a confirmation prompt.

Every row now carries the same two controls, whether the access was granted here or inherited: one verb menu — Owner, Can edit, Can read, Can download, then Deny — and the same Remove (x) beside it. Picking a set writes the difference in the background: verb-scoped `deny` entries at this target for verbs the set drops that a parent still grants, a plain revoke for verbs granted only here, and grants for anything the set adds that nothing confers. Picking a set at or above the parent's lifts local denials the new set no longer needs. Deny blocks every verb, including read, at this target — the row stays, reading "Denied here", with the same menu ready to lift it.

The underlying bug was in the payload: `GET /api/workspace/:id/access` reported only grants, so a principal whose only local entry was a denial had no local grant, fell into the collapsed parent section, and read as removed (and a principal denied every verb held nothing at all, so the row vanished outright). The response now also carries `denials` (per principal, per verb, with the same `direct`/`ancestor` source split as `sources`) and `deniedHere` (the principals this target restricts). The dialog decides its sections by local entry — grant or denial — not by local grants alone.

Remove (x) is unchanged: on a direct row it revokes here; on an inherited row it still asks whether to remove at the parent or restrict just this item.

## Testing
- Backend: 255 files / 4083 passed / 1 skipped / 0 failed
- Frontend: 174 files / 2388 passed / 0 failed
- Booted from source and verified all 8 acceptance criteria live (see ticket's Local Testing Evidence)

---
## Summary by cubic
Restricting someone below what a parent folder grants no longer makes them disappear from the Manage access dialog, and no longer detours through a confirmation prompt.

- Every row now shows the same verb menu (Owner, Can edit, Can read, Can download, Deny) and Remove, whether access is granted here or inherited.
- Picking a menu item states the whole verb set and writes the difference in the background: verb-scoped `deny` entries for verbs a parent still grants, revokes for grants only here, and grants for verbs nothing confers.
- Deny blocks every verb here; the row stays, reading "Denied here", and the menu can lift it.
- The access view now reports `denials` and `deniedHere`, so rows are placed by local entry (grant or denial) instead of local grants alone.
- Remove is unchanged: direct rows revoke here; inherited rows still ask whether to remove at the parent or restrict just this item.
- `deniedHere` now counts only local denies that actually restrict (unknown-role denies in a file's frontmatter are dropped), and the Remove prompt offers only parent folders that grant the principal.

<sup>Written for commit a7a144c453efbb8baa08d3666a66ea82820fb13a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Bevel-Software/Hexis/pull/268?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


h3xxit:
Addressed the four review findings in a7a144c4. Three taken as written; one resolved in the opposite direction, explained below.

**P2 — unknown-role denial in file frontmatter (`access-control.service.ts`).** Valid, fixed by the second option you offered: don't report what isn't enforced. `loadModel` strips unknown-role entries from every `access.md`, and `resolveScopes` filters them out of a file's own frontmatter (the one scope it does not pre-filter), so such a line denies nothing and `denialSources` is right to report nothing. `locallyDeniedPrincipals` now skips them on the same `roleKnown` rule, so `deniedHere` and `denials` name the same principals. Preserving the row instead would have put "Denied here" on a restriction the gate does not apply, against the ticket's own acceptance criterion that the shown verdicts match what the server enforces. Two tests added (`access-restrict-here.test.ts`): a known group token in file frontmatter is reported in both halves *and* enforced; an unknown one is reported in neither.

**P3 — `deniedHere` docstring (`access.routes.ts`).** Correct, and the tests you cite are the contract. Reworded: `deniedHere` is every principal a deny on this target names, however many verbs it took; a partial restriction leaves them in the eligible lists for the verbs they keep, and only a principal denied every verb drops out of those lists. Also noted that an unknown-role deny appears in neither field.

**P3 — brittle `textContent` assertion.** Agreed. The menu-order check now matches items by accessible name and compares the elements, with the provenance notes asserted separately as the `aria-description`s they are.

**P2 — `lookupDenials` in the post-revoke Remove chain (`ManageAccessDialog.tsx:1268`).** I've gone the other way, because the suggested line makes the dialog offer an act that increases access in answer to a click asking to remove it. That chain exists to finish a half-removal: the direct entry is gone, the person *still has access from a parent*, so we offer "Remove from &lt;parent&gt;", which calls `doRemoveFromParent` → a revoke at that ancestor. Feed it a denial ancestor and the primary button deletes the parent's **restriction** — the person gains access from an x that asked for less. And in the case you describe (direct deny + inherited deny, no grant above), the removal is already complete after the local splice: the parent still denies, so there is nothing left to finish and nothing to prompt about.

The real inconsistency you found is one row up, at the other call site: `classifyManage` puts denial ancestors into `row.ancestors`, and the x on an inherited row hands exactly that list to the same confirm dialog — so a row whose only ancestor entry is a parent's denial (reachable when the person reads through a role) did offer "Remove from Sales" for a folder that only restricts them. Fixed by splitting the two uses: `ancestors` stays every folder holding an entry (it is what "via Sales" and the parent-folder grouping report), and a new `grantAncestors` — grants only — is what the Remove flow acts on. All three `setConfirmRemove` sites now read grants alone, and the post-revoke chain says why in a comment. A row a parent merely restricts now falls to the no-ancestor branch of the confirm: restrict-here, and nothing else. Covered by a new test in `ManageAccessDialog.restrict.test.tsx`.

Backend 4085 pass, frontend 2392 pass; `core-backend` / `core-frontend` typecheck clean (`apps/server` needs the workspace packages built first — pre-existing here, unrelated to this branch).

PR #270, opened 2026-09-19
Saved-Key-Is-Probed — Every place that saves a tool key probes it afterwards and says when the provider rejected it

h3xxit:
Ticket: `Data/Engineering/Knowledge/Hexis/Tickets/Saved-Key-Is-Probed.md`

## Summary
After a successful save on the tool page, the Connect your tools page, and the vault, the same health-check probe now runs and its result is shown next to the saved key:

- A rejected key shows the provider's status text, as the tool page already did.
- A tool whose manual defines no health check shows "Unverified" with a one-line explanation.
- Saving is never blocked by the probe — the value is stored first, then the probe runs.
- The submitted secret never appears in the result text, logs, or network errors.

Introduces a shared `packages/core-frontend/src/modules/secrets-vault/probe/` module (`useSavedKeyProbe`, `probe-verdict`, `SavedKeyProbeResult`) used by `ToolConnectionSection`, `ConnectToolsPage`, and `ToolSecretsPanel`/`SecretsPage`, so all three surfaces share one probe result presentation.

## Test plan
- `pnpm typecheck` clean across affected packages.
- `pnpm test`: 6783 passed, 0 failed (repo-wide); component suites cover accepted, rejected, and unverifiable keys on each surface.
- Live-booted verification (docker compose build) against real/fake GitHub PATs and a health-check-less manual: confirmed accepted/rejected/unverifiable verdicts, save-not-blocked-by-probe, and no secret leakage in responses or logs.

---
## Summary by cubic
All three surfaces that save a tool key — the tool page, Connect your tools, and the vault — now run the same health-check probe after a save and show the verdict next to the saved key. Previously only the tool page probed; the other two stored a rejected key silently, making it look identical to a working one.

**New Features**
- Rejected keys show the provider's status text in a danger banner.
- Tools with no health check show `Unverified` with a one-line explanation, including on the tool page.
- Probe failures say the check could not run, never that the key is wrong.
- The value is stored and the field cleared before the probe starts, so a hanging or failing probe never blocks saving.

**Bug Fixes**
- Submitted secrets never appear in the result or rendered text; provider echoes of short credentials are redacted too, and redaction leaves the rest of the provider's message readable.
- Probe answers are keyed to the tool and its definition revision, so swapping tools can't leave a stale `Connected` verdict.
- Save and refresh refetches are quiet and newest-wins, so an older response can't unmount the verdict or overwrite a newer one.
- Blank probe details fall back to a generic message instead of rendering an empty word.

<sup>Written for commit c29db2b1ece1502232db5790e314190966553d90. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Bevel-Software/Hexis/pull/270?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


h3xxit:
All seven cubic findings on `1695061d` are addressed in code — none needed a disagreement. Branch is also merged up to `origin/dev` (`b1489e40`, clean merge). New head: `226163f5`.

**P1 — a credential shorter than six characters survived redaction** (`probe-verdict.ts:88`)

Valid, and the root cause was on the server rather than at the hint: `SecretRedactor.remember` had a six-character floor, so a four-character key the provider echoed back reached the page verbatim — readable by anyone who can probe the tool, which is anyone who can read it. The floor is gone; every non-blank resolved value is remembered now. What the length decides is how a value may **match**: below six characters it is replaced only where it stands alone (bounded by something other than a letter or digit), because `substitute` reports every `${VAR}` a manual resolved and not only the credential ones — a `${VERSION}` of `1` blanked wherever its character fell would turn `answered 401` into `answered 40[redacted]` and destroy the message the quote exists to carry. Two tests, both of which fail against the old floor.

**P2 — a previous tool's verdict beside a new tool's variable** (`ToolSecretsPanel.tsx:47`)

Valid: `ToolRenderer` mounts one panel and swaps `tool` as the reader opens a different `.tool` file. Fixed in the hook rather than at the caller, so no host has to remember to key it — the slug is now half the identity of every answer `useSavedKeyProbe` holds, alongside the revision stamp, and is compared at render exactly as the stamp already was (race-free, no clearing effect). New `ToolSecretsPanel.test.tsx` covers the swap and fails if the comparison drops the slug.

**P2 — an older `listToolSecrets` overwriting a newer one** (`SecretsPage.tsx:175`)

Valid. Both pages carry a monotonic `loadSeq` now; only the newest load may apply its lists, raise an error, or clear the spinner. Applied to `ConnectToolsPage` too — it had the identical race through `getConnectPending`.

**P2 — `Refresh` discarding a verdict** (`ConnectToolsPage.tsx:84`)

Valid. The header button is quiet on both pages; the feedback is the disabled state rather than a spinner over a list that is already on screen.

**P2 — `stores the value first` did not pin the order it named**

Correct, and the suggested shape is what it does now: `setAdminVar` is held unresolved, the test asserts `checkToolConnection` has **not** been called and the editor is still open, then releases and asserts the probe starts. Verified by moving `onSaved()` above `await action()` in `ToolVarRow` — the old test passed, the new one fails.

**P3 — a blank `detail` reaching the page**

Taken. `probeWords` treats blank and whitespace-only detail as missing before choosing a fallback. The untrimmed value is what a non-blank detail renders, so the provider's words still arrive exactly as the server composed them.

**P3 — the password-field claim in `never repeats the submitted secret`**

Correct on the mechanism: a controlled input's value lives on the DOM property, not in `textContent` or serialized `innerHTML`. The field is asserted directly now, with the document-level checks kept for everything else. Verified by deleting `setValue('')` in `KeyRow` — the two document checks pass, the field assertion fails.

---

Evidence on the new head: `pnpm typecheck` clean across all 7 projects; full `pnpm test` green (256 backend files / 4096 tests, 176 frontend files / 2420 tests, 0 failures); `pnpm ds:check` no regressions. The touched files produce the same seven pre-existing lint errors as `HEAD` did — none added, checked by stash-and-compare.

Still open from Local Testing and untouched here: `connection-probe.service.ts:422` and `:437` return `err.message` without the redactor. Both fire before substitution, so the coder's read is that no secret can reach them; nothing in this pass confirms or refutes that, so it stays a review question.

h3xxit:
Second cubic run (`05ee1be1`, against `226163f5`) raised three findings. All three are addressed in code — none needed a disagreement, though one is worth a correction on the stated symptom. New head: `c29db2b1`. Base still merged up to `origin/dev` (`b1489e40`).

**P2 — `standingAlone()` can redact ordinary punctuation** (`connection-probe.service.ts:224`)

Valid, and worse than the comment implies — this was my own regression from the previous round. A value made *entirely* of punctuation hit neither branch, so `standingAlone('.')` compiled to `/\./g` and replaced every sentence period and every dot of every hostname in the provider's quote. A `${SEPARATOR}` of `.` or a `${PATH_PREFIX}` of `/` is an ordinary thing for a manual to declare, so this was reachable.

Taken as suggested: the word-character guard now goes on both ends unconditionally. I want to be explicit about what that costs, since my original asymmetry was a deliberate (wrong) trade: a sub-six-character value whose edge is punctuation and which the provider echoes *glued to a word character* (`sk-ab` for a value of `-ab`) is no longer redacted. That is the right way round — the miss needs a credential short enough to carry almost no secret, while over-redaction destroys the quote for everyone and cannot be recovered from by the reader. Values of six characters or more never reach this path.

The three live-verified behaviours from Local Testing all survive the change: `rest` between `/` and `"` still redacts, `doc` inside `docs` still does not, and the 4-character-credential case from the first round still redacts. New test `leaves ordinary punctuation alone when a resolved value IS punctuation` fails against the old asymmetric version.

**P3 — the `Refresh` test locked in a contradictory row** (`SecretsPage.test.tsx:392`)

Valid. Feeding the pre-save list back on every load left the badge reading `Needs a key` under a verdict reading `Connected` — a state the real flow never produces, since a save that landed makes the next list report `userConfigured: true`. Every load after the first now says so, and the test additionally asserts the row is coherent (`Set` above `Connected`) and that `checkToolConnection` is still called exactly once, so a refresh is not silently re-asking the provider. Applied to the `ConnectToolsPage` sibling too, which had the identical fixture. Both still fail if the header `Refresh` is made loud again — re-verified after the rework, since changing a fixture can quietly defang a test.

**P3 — `releaseStore()` outside `act()`** (`ToolConnectionSection.test.tsx:627`)

Applied — it matches the convention the definition-edit race test in the same file already follows, and making the flush explicit rather than incidental to the two `waitFor`s that follow is a genuine improvement for an *ordering* test.

One correction to the finding's stated symptom, for the record: the predicted `not wrapped in act(...)` warning does not actually occur. I measured it both ways — `grep -ci 'not wrapped in act'` over the test's output returns **0** with the wrap and **0** without it. So this was a robustness/consistency change, not a fix for an observed warning.

---

Evidence on `c29db2b1`: `pnpm typecheck` clean across all 7 projects (after rebuilding `shared` + `core-backend`); full `pnpm test` exit 0 — core-backend 256 files / 4097 passed + 1 skipped, core-frontend 176 files / 2420 passed, shared 187, mcp-core 111, zero failures; `pnpm ds:check` 206 vs baseline 215, no regressions; the touched files produce the same seven pre-existing lint errors as before, none added.

Still open and still untouched: `connection-probe.service.ts:422` and `:437` return `err.message` without the redactor. Both fire before substitution, so the coder's read is that no secret can reach them; neither this pass nor the live run confirmed or refuted it.

PR #272, opened 2026-09-19
Session-First-Call-Reproduced — first start_session call reproduced under load

h3xxit:
Ticket: `Data/Engineering/Knowledge/Hexis/Tickets/Session-First-Call-Reproduced.md`

## Summary

- Built a load probe (`packages/core-backend/src/modules/mcp/first-call-probe.ts` + CLI) that opens fifty fresh connections and issues one `start_session` call each, recording per-call timings and failures.
- Ran it against the full platform (50 fresh connections, one first `start_session` call each): 50/50 succeeded, 50 distinct ids, 0 failures. No platform-side failure reproduced.
- Confirmed the reported error id shape (`req_011...`) is an Anthropic API request id — nothing in `packages/` or `apps/` mints ids in that shape, so the failure was raised between the model and the tool, not on our side.
- Since no platform-side bug reproduced, updated the `start_session` tool description instead: a failed call created nothing and can be safely retried; a retry landing after a success returns a new (harmless) id.
- Verified locally by building the branch from source (`deployment/docker-compose.build.yml`) and re-running the fifty-connection probe twice (cold + steady state) plus a tool-mode run: all 50/50 success, 0 failures, 50 distinct ids each time. Also confirmed live over MCP `tools/list` that the new description text is present, and demonstrated that two `start_session` calls on one connection return distinct ids while the first id keeps working.

## Tests

- `packages/core-backend/src/modules/mcp/__tests__/first-call-probe.test.ts` (19 tests)
- `mcp.e2e.test.ts`: fifty-fresh-connections reproduction case
- `workspace.tools.test.ts`: `start_session` load, retry-safety, and description assertions
- `pnpm --filter @bevel-software/platform-core-backend test`: 257 files / 4118 passed / 1 skipped / 0 failed
- `pnpm typecheck` clean; eslint clean on touched files

## Changeset

`.changeset/start-session-retry-is-safe.md` (patch, `@bevel-software/platform-core-backend`)

---
## Summary by cubic
Adds the first-call load probe used to investigate the ticket and updates the `start_session` tool description so callers know retrying is safe: a failed call created nothing, and a retry that lands after a success returns a new, harmless id. The reported failure did not reproduce on the platform side — 50 fresh connections, one first call each, all succeeded — and the error id (`req_011...`) is an Anthropic API request id, so the failure happened between the model and the tool, not in this codebase.

**New Features**
- `probe:first-call` opens 50 fresh connections, issues one `start_session` call on each, and prints timings plus every failure with the platform's own error text.
- `--mode tool` targets the raw tool route without the MCP transport, so a future failure can be placed on one side of it.
- The probe refuses input it cannot honor before opening anything and escapes every value it prints — refusals included. A mistyped flag no longer silently runs the default probe, and a bare `--bearer` no longer becomes the literal bearer `true` and reads as fifty platform failures.
- The e2e suite runs the probe against the full platform and pins 50 distinct ids with no failures.

<sup>Written for commit 998ec9dc490329201c72b6657aa667535c9e8f21. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Bevel-Software/Hexis/pull/272?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


h3xxit:
All eight cubic findings accepted and fixed in `cd945fa1`. No disagreements.

| Finding | What changed |
| --- | --- |
| `first-call-probe.ts:195` — unescaped log interpolation | `formatProbeReport` now puts `f.error`, `report.target` and `report.toolName` through the shared `printable`. New test asserts a report cannot gain a line, or an ANSI escape, from what a remote endpoint sent. |
| `first-call-probe.ts:237` — `new URL()` throws inside the pool worker | `normalizeBaseUrl` validates once in `probeFirstCall`, before the burst, and rejects a non-http(s) scheme too. A bad URL is a config error (nothing was attempted), not fifty lost results — both modes now answer it identically. |
| `first-call-probe.cli.ts:52` — `--concurrency 0` silently clamped to 1 | Every numeric flag is a count, parsed by one `count()` helper that refuses non-integers and anything below 1. `probeFirstCall` refuses the same values for a programmatic caller, so the clamp can no longer relabel a serial run as a burst. |
| `first-call-probe.cli.ts:88` — unhandled rejection on a bad flag | `main()` is wrapped. A `ConfigError` (including the ones `probeFirstCall` raises) prints the sentence plus the usage block and exits 2, matching the missing-`--base-url` path; a genuine crash still gets its stack and exit 1 rather than a usage block implying a typo. Verified: `--concurrency 0`, `--connections abc`, `--base-url not-a-url` and a missing `--base-url` all exit 2. |
| `first-call-probe.test.ts:172` — the "never answers" test measured a refusal | Renamed to say it measures a refused connection, and a real one added: a `net` server that accepts the socket and answers nothing. It exercises the connect-timeout branch for the first time (test takes ~1.0s against a 1s ceiling). |
| `mcp.e2e.test.ts:644` — 60s budget under a 2×30s worst case | Budget raised to 120s **and** `timeoutMs: 20_000` passed, so in the flaky-first-call scenario the test exists to catch, the failure list and the report actually get printed instead of a bare vitest timeout. |
| `workspace.tools.test.ts:1514` — spy leak on a rejected request | `errorLog.mockRestore()` moved into a `finally`. |
| `eslint.config.js:34` — comment claims the CLI is not deployed | Correct: the package tsconfig includes all of `src`, so it compiles into `dist` and is copied into the image. Reworded to the invariant that holds — no deployment *executes* it — with the reason spelled out, rather than excluding `*.cli.ts` from the emit and losing that coverage. |

`pnpm typecheck` clean; core-backend suite 257 files / 4121 passed / 1 skipped / 0 failed; eslint clean on every touched file.

h3xxit:
Round two: one new finding on `cd945fa1`, accepted and fixed in `82f39414`. The other eight comments on this PR are the round-one findings, all fixed in `cd945fa1` (GitHub re-anchors them to the current head when the lines still exist — they are the same comment ids, not new).

**`first-call-probe.cli.ts:62` — argv copied verbatim into a `ConfigError`.** Correct, and it is the same rule I applied to the report block but not to the CLI's own refusals, which print to the same terminal and the same CI log. Every argv value composed into a `ConfigError` now goes through `printable`. Before:

```
$ probe:first-call --base-url http://127.0.0.1:1 --bearer x --connections $'9\n  50/50 succeeded, 0 failed'
--connections must be a positive integer, got 9
  50/50 succeeded, 0 failed
```

After — one line, escaped rather than dropped, so the operator still sees exactly what was typed:

```
--connections must be a positive integer, got "9\n  50/50 succeeded, 0 failed"
```

Argument handling moved to `first-call-probe.cli-args.ts` in the same commit, because the refusals could not be tested where they were: the entrypoint is a top-level-`await` script, so importing it runs a probe, and every check on it so far had been by hand. The `.cli.ts` keeps only what a suite cannot reach — argv, printing, exit code — and 18 new tests cover the rest, including that a refusal stays one line whatever is put in the flag.

Verified: `pnpm typecheck` clean; core-backend suite 258 files / 4139 passed / 1 skipped / 0 failed; eslint clean on the three touched files; `--concurrency 0`, `--base-url not-a-url`, `--mode grpc` and the forged-newline case above all exit 2 with a single-line reason, `--help` exits 0.

h3xxit:
Round three: two new findings on `82f39414`, both accepted and fixed in `998ec9dc`. Everything older on this PR is round one (`cd945fa1`) and round two (`82f39414`), already fixed — GitHub re-anchors those comments to the current head when the lines still exist, so their `commit_id` moves even though the ids and timestamps do not.

Both findings are the same failure mode, and for this file it is the worst one there is: a probe run is evidence that gets pasted into a ticket log, so a command that quietly runs something other than what was typed produces a confident wrong number.

**`cli-args.ts:43` — a mistyped flag was ignored.** `--concurency 1` was dropped and the default fifty-wide burst ran, reported as the one-at-a-time run the operator thought they had asked for.

**`cli-args.ts:45` — a value flag with no value became `"true"`.** `--bearer` alone probed with the literal bearer `true`, so all fifty attempts came back 401 and the report read as fifty platform failures — a reproduction of a bug that was never there. Given what this ticket is (deciding whether a failure is ours), that one could have sent the whole investigation the wrong way.

The flags are now a table. Anything not on it is an unknown option, a value flag without its value is named, and a stray word that belongs to no flag is refused too. The usage text is held to the same table by a test, so a flag cannot exist undocumented.

One thing the table had to keep accepting, which the suggested patch would have broken: pnpm passes the end-of-options `--` straight through. The documented invocation reaches argv as `tsx … first-call-probe.cli.ts -- --base-url …`, verified by running it, so `--` is skipped rather than refused — with a test pinning that.

Live against the entrypoint, all exit 2 on one line:

```
--concurency 1                  → unknown option "--concurency"
--bearer        (no value)      → --bearer needs a value
--tool          (no value)      → --tool needs a value
junk                            → unexpected argument "junk" — every input to this command is a --flag
$'--nope\nFORGED: 50/50 …'      → unknown option "--nope\nFORGED: 50/50 succeeded"
```

and the documented pnpm form still runs a real burst (exit 1 on attempt failures, as designed).

Verified: `pnpm typecheck` clean; core-backend suite 258 files / 4146 passed / 1 skipped / 0 failed (25 in the CLI-args suite, up from 18); eslint clean on the touched files.

PR #275, opened 2026-09-19
Setup-Ends-With-Connector — The Claude setup instructions end with the connector

h3xxit:
Ticket: `Data/Engineering/Knowledge/Hexis/Tickets/Setup-Ends-With-Connector.md`

## Summary
- Converts the admin-facing marketplace registration instructions (Deployment settings → Marketplace) into a carousel using the same shell as the person-facing carousel: progress strip, Back / Next / Review again, arrow-key and Home/End handling, one screenshot mounted at a time. The four existing shots keep their order and alt text.
- Restricts the generated-credentials card to mount only when the drawer is open AND the Add-configuration slide is active.
- Adds a shared connector step, rendered identically at the end of both carousels (person-facing after "Add Hexis all", admin after "pick this deployment as the GitHub instance"), driven by one slide definition so the copy can't drift between the two.
- Wires two new screenshots for the connector step through `claude-setup-shots.ts` with alt text and percentage callout boxes, consistent with the existing nine shots.
- Updates `docs/claude-cowork.md` with the same connector step under "Add the marketplace (every person)", including the verification line and a troubleshooting paragraph, and corrects the "five-screen walkthrough" sentence to match the current page.
- Adds tests for: the connector slide as the last slide of both carousels (naming Add for your team), the admin carousel's shell (progress strip, step counter, Review again at the end), the credentials card mounting only on its own slide, and the two new shots' alt text/callout boxes.
- No backend change, no change to the marketplace compiler, the `Add to Claude` install link, or registration mechanics.

## Notes for reviewers
- The two new screenshots (`10-connector-not-added.webp`, `11-add-custom-connector.webp`) are placeholder renderings, not captures of a real deployment. The asset README marks them as such and documents the re-shoot recipe. The prepared originals referenced in the ticket Notes (1280x316 and 1056x866) were not available in the working tree at implementation time — a reviewer should decide whether to ship these placeholders or wait for the real captures before merging.

## Test plan
- `NODE_ENV=test pnpm install && pnpm build && pnpm typecheck && pnpm test` — 258 files, 4146 passed / 1 skipped, exit 0; build and typecheck clean.
- Change touches only frontend components, docs, and static assets — no runtime behavior — so no docker compose boot was required.

---
## Summary by cubic
Both Claude setup walkthroughs (person-facing and admin registration in Deployment → Marketplace) now end with the connector step, so installing Hexis all no longer looks complete while the plugin's MCP server is still disconnected. The admin registration flow now uses the same carousel shell as the person-facing tutorial, and the shared connector slide is defined once so the two copies can't drift.

**Reviewer notes**

- Generated credentials mount only when the drawer is open and the Add-configuration slide is active.
- The two connector screenshots are placeholder renderings; a shot-level `illustration` flag prints a line under each frame so readers know the detail may differ.
- Arrow keys that would not move the carousel are left to the browser instead of swallowed.
- No backend, marketplace compiler, or registration mechanics changed; docs, tests, and a changeset are updated to match.

<sup>Written for commit 7257437fd94375281e98d533bbe81887476e2cd8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Bevel-Software/Hexis/pull/275?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


h3xxit:
All six findings are addressed in `54074aa2`, on top of a merge of `origin/dev` (clean, no conflicts). Thread by thread, since cubic reads the head rather than the replies:

**P2 — two drawings ship into a published package's setup carousel** (`claude-setup-shots.ts:127`)

Taken, and answered a third way. Both options offered remove something: waiting for real captures blocks the step this ticket exists to add, and gating the connector slide behind the swap ships a walkthrough that still stops one step short of a connected knowledge base — the exact failure the step is for.

The real harm you name is a reader holding their own Claude beside a drawing and concluding they are on the wrong screen. That is now said on the page, not only in the asset README. `Shot` gained an optional `illustration` flag; a shot carrying it prints a line under its frame — *"Illustration, not a capture: this is drawn to the screen's layout, so the detail on your Claude may differ. The instruction above names the row and the control to look for."* The nine real captures declare nothing and print nothing, which is what stops the note becoming decoration on every screenshot.

So the PR's claims are confined to the alt text and the instruction, as you asked — the difference is that the page now says that out loud instead of the reader having to infer it. The README says a capture means deleting the flag *and* re-measuring the boxes, in those terms, so the misalignment you flag cannot be half-fixed. Tests cover both the flag on the shots and the note appearing on the connector slide and nowhere else.

**P3 — key handler swallows keys that cannot move it** (`SetupCarousel.tsx:50`)

Correct, and it matters more than the P3 suggests because the section governs both routes now. `handleKeys` computes the slide it would move to and returns early — no `preventDefault`, no `setState` — when that is the slide already on screen. Left on the first, Right on the last, Home at the start and End at the end all go back to the browser. The admin tests assert both directions through `fireEvent`'s return value, which is `false` exactly when `preventDefault` was called, so the no-op edges are now a guarded contract rather than a side effect.

I did not take the second half ("handle only when focus is on the progress/toolbar controls"). Arrow keys on a `role="region"` carousel are conventionally the region's, and a reader who tabbed to a screenshot link is still inside the carousel; narrowing it to the toolbar would make the keys work in some parts of the widget and not others. The scroll complaint is fully answered by not preventing what we do not use.

**P3 — region name duplicates the drawer's** (`ClaudeRegistrationSteps.tsx:122`)

Taken. `label="Claude registration steps"`. The `<summary>` keeps "Register this deployment with Claude" and the eyebrow inside still announces the position, so nothing is lost and the name is no longer said twice on the way in.

**P3 — admin tests only simulate End** (`MarketplaceSection.test.tsx:135`)

Taken. Two tests added: one walks ArrowRight/ArrowLeft/Home/End including both no-op edges, and one asserts "Review again" restarts at step 1 with the counter back to `1 / 5`, Back disabled, and `aria-current="step"` back on the first progress button.

**P3 — containment test guards only the far edge** (`claude-setup-shots.test.ts:58`)

Taken, as suggested. A negative offset leaves the image and satisfied `x + w <= 100`, so the test did not enforce what its comment promised.

**P3 — "the registration half is the same walkthrough"** (`docs/claude-cowork.md:93`)

Taken, as suggested. It reads "The registration half lives in **Deployment → Marketplace** and ends on the same connector step." The registration carousel is five slides sharing only the last; the old wording said otherwise.

**Also in this push:** the branch was missing a changeset. Added.

Full suite on the new head: 273 files, 4404 passed / 1 skipped, exit 0; build and typecheck clean.

h3xxit:
Correction to the count above: there are **seven** findings, not six, and the one my rundown skipped is

**P3 — "the five actions every person takes" in the comment above the walkthrough test** (`ExternalAgentAccessPage.test.tsx:470`)

Taken. Fixed in `54074aa2` alongside the rest — the personal carousel is six steps since the connector was added after "choose the Hexis all row", and that comment was the occurrence the PR description's "five-screen walkthrough" pass missed:

```diff
-   * the five actions every person takes, one at a time, and none of the
+   * the six actions every person takes, one at a time, and none of the
```

All seven threads are addressed in code on the current head.

h3xxit:
New head `7257437f`, on top of a merge of `origin/dev` (`fe70dc1e` → `b4e77729`, clean).

**P3 — the added End keydown is dead** (`ExternalAgentAccessPage.test.tsx:548`)

Taken — and the dead keypress was the smaller half of it. Worth spelling out, because the second half is a hole the suggested edit would have left open.

That block sits in front of an older assertion: *"Home goes back to the first, so End is not a one-way door."* That only means anything when the carousel is on the **last** slide at the moment Home is pressed. The `Home` I spliced in above it to check a real-capture slide had already made the round trip's return leg, so the final `Home` was pressing Home on slide 1 — and the assertion would have passed on a handler that had stopped responding to `End` entirely. The dead `End` was me half-noticing that and trying to put the state back.

So rather than just deleting the `End` (which leaves the doubled `Home` and the hollowed assertion), the two presses collapse into one that does both jobs:

```js
expect(within(carousel).getAllByText(/Illustration, not a capture/)).toHaveLength(2);

// And Home goes back to the first, so End is not a one-way door. The
// first slide is a real capture, so the same press is what shows the
// note is per-shot rather than part of the shell.
fireEvent.keyDown(carousel, { key: 'Home' });
expect(within(carousel).getByRole('group')).toHaveAccessibleName(
  'Step 1 of 6: Select this deployment in Claude Code',
);
expect(within(carousel).queryByText(/Illustration, not a capture/)).toBeNull();
```

One press, no dead statement, and the End→Home round trip is load-bearing again. Both directions of the note stay guarded: drop the `illustration` flag and the `toHaveLength(2)` fails; render the note on every shot and the `toBeNull()` fails.

**Still open, unchanged:** the P2 on the two placeholder screenshots (`r4052636859`). Answered in [the earlier comment](https://github.com/Bevel-Software/Hexis/pull/275#issuecomment-5749820874) — neither option offered was taken, because both remove the connector step this PR exists to add; instead the page now tells the reader it is looking at an illustration. That one is a human call, not something a further commit settles.

`r4052636887` (the `preventDefault` finding) carries no ✅ tick, but it is fixed in `54074aa2` — `handleKeys` returns before any `preventDefault` when the target slide is the current one. GitHub could not re-anchor the comment (its `commit_id` is still the original head, `line: null`), so cubic had nothing to re-verify against.

Full suite on the new head: 276 files, 4469 passed / 1 skipped, exit 0; build and typecheck clean.

PR #276, opened 2026-09-19
Skill-Lifecycle-Removed — a skill's lifecycle and owner are no longer the platform's business

h3xxit:
Ticket: `Data/Engineering/Knowledge/Hexis/Tickets/Skill-Lifecycle-Removed.md`

## Summary
- Removes `owner` and `lifecycle` from the skill contract, frontmatter parser, pending-skills projection, and frontend item types.
- The compiled marketplace now includes every readable skill regardless of frontmatter; the retired filter and the `retired-skills` link refusal (plus its repair path) are gone.
- The library, link picker, and skill page's shared-via section no longer show or filter on Deprecated/Retired.
- The agent guide's skill metadata section drops `lifecycle`/`owner` docs, keeping `version`.
- A SKILL.md still carrying `metadata.lifecycle`/`metadata.owner` is read like any other skill — the keys are ignored, not rejected.
- Replaces the retired-behaviour tests with one asserting a skill with `lifecycle: retired` compiles into the marketplace and is listed over MCP like any other.

## Test plan
- `NODE_ENV=test pnpm test` — 456 files, 6830 passed + 1 skipped, 0 failures.
- Booted commit from source via `deployment/docker-compose.build.yml`; verified all acceptance criteria against the running app with a seeded SKILL.md carrying `metadata.lifecycle: retired` + `metadata.owner`.

---
## Summary by cubic
Implements the Skill-Lifecycle-Removed ticket by dropping `owner` and `lifecycle` from the skill contract and frontend, so the platform no longer governs a skill's lifecycle. `SKILL.md` files that still carry `metadata.lifecycle` or `metadata.owner` are read normally; those keys are ignored, not rejected.

**Behavior changes**
- The marketplace now ships every readable skill; `retired` no longer keeps one out.
- Linking a root no longer refuses retired skills, so the `retired-skills` error and its "link them one by one" path are gone.
- The library, link picker, and shared-via panel no longer show Deprecated/Retired badges or filter on lifecycle.
- `metadata.version` remains the only skill metadata the agent guide documents.

**Tests**
- Backend: a `lifecycle: retired` skill compiles into the marketplace and is listed over MCP.
- Frontend: catalog mapping drops `lifecycle`/`owner` from older payloads, the library card badges nothing, and the link picker and shared-via panel still offer retired skills.

<sup>Written for commit f49e77e5791dbaeb0089b1fe99faba09f85a28ac. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Bevel-Software/Hexis/pull/276?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


h3xxit:
> **question** — Missing evidence: PR claims multiple behavior changes, but new tests pin only backend catalog/linking behavior; frontend lifecycle badge/filter removals are untested in this diff.

Agreed, and fixed. Checking first: no frontend test had *ever* named `lifecycle` — `grep -rn 'lifecycle\|retired\|Deprecated' packages/core-frontend/src --include='*.test.*'` returns only unrelated prose — so the three frontend removals were unpinned before this branch as well as after it. Added one test per removal:

| Removal | Test |
| --- | --- |
| Deprecated/Retired badge | `LibraryCard.test.tsx` — “badges no lifecycle, whatever the frontmatter still says”, over `deprecated`/`retired`/`active` |
| Link picker's `lifecycle !== 'retired'` filter | `LinkSkillPanel.test.tsx` — “offers a skill whose frontmatter still says retired, like any other” |
| Shared-via section's `retired ? [] : …` | new `SharedViaPlugins.test.tsx` — offers every managed plugin to that same skill, and withholds one only for a membership it already has |

The two picker tests carry `lifecycle: 'retired'` on the item through a deliberate cast: the key is gone from `LibraryItem`, so the case under test is a server that still sends it changing nothing in the UI.

Each one was run against the pre-change component (`git checkout origin/dev --` the four frontend files) and fails there — 3 failed / 21 passed — so none is vacuous. On the branch: 40 files / 595 tests green in `src/modules/library`, 15 files / 198 tests green in core-backend `src/modules/plugins`, `pnpm typecheck` clean.

One extra tidy in the same commit: `repair()` still called `await this.resolvedSkills(root)` and discarded the result, under a comment naming “the same retired-skill rule as `link`”. That rule is gone, so the call and its comment are too.

h3xxit:
Re [cubic's P3 on `LibraryCard.test.tsx:109`](https://github.com/Bevel-Software/Hexis/pull/276#discussion_r4053266747) — agreed, and taken.

The criticism is correct: `LibraryCardProps` has no `lifecycle` key, so the cast was doing the work, and what the assertions actually proved is that React ignores an undeclared JSX prop. It would still go red if a badge rendering those exact words came back — which is why it isn't deleted outright — but it exercised none of the frontmatter→catalog→item→card path its title claimed, and a test whose name overstates its reach is worse than no test.

Took the first of the two suggestions, as the stronger one. New `library-data.mapping.test.tsx` pins the removal at the reachable layer:

- a `LibrarySkillSummary` still carrying `lifecycle: 'retired'` and `owner: 'gtm-team'` — an older server mid-upgrade, or a `SKILL.md` still declaring the keys — maps to a `LibraryItem` with **neither** key, asserted with `in` rather than truthiness, while `name`/`version`/`path` survive intact;
- that item is `toEqual` the same skill sent without the keys, which is the whole claim: the platform has no opinion left to form about a skill's lifecycle.

That is the load-bearing one. Once the mapper drops the keys there is nothing downstream for a card, picker or panel to badge or filter on.

The card test stays, reduced to the narrower thing it honestly is — a re-introduction guard on the rendering — with the three-value loop dropped, retitled “renders no lifecycle badge, even when handed the removed prop”, and a comment scoping it explicitly against the mapping test. Second suggestion (drop the cast too) I did not take: without it the assertion passes against the old component as well, since the old badge only rendered when a `lifecycle` prop was present, so it would become genuinely vacuous rather than merely narrow.

Both new mapping tests fail against the pre-change mapper (`git checkout fa45be58 -- library-data.tsx library.api.ts` → 2 failed / 2). Full suite on the merged head: 460 files, 6868 passed + 1 skipped, 0 failures; `pnpm typecheck` clean. The branch is also merged up to `origin/dev` at `5fe7e48b` — `plugin-page-parts.tsx` auto-merged and was checked to confirm the `lifecycle` prop did not come back with it.

PR #277, opened 2026-09-19
Sidebar-Footer-Aligned — The sidebar's bottom-left rows sit on the sidebar's own grid

h3xxit:
Ticket: `Data/Engineering/Knowledge/Hexis/Tickets/Sidebar-Footer-Aligned.md`

## Summary
Aligns the sidebar's bottom-left rows (the setup reminder and the CHANGE REQUESTS row) to the same left inset and vertical rhythm as the tree rows above them.

- Introduced `IntegrationsSetupReminder.tsx`, a truncating reminder line with a tooltip on narrow widths and an always-visible "Finish now" link.
- Reworked `PullRequestsForMe.tsx`'s CHANGE REQUESTS row so its label truncates and its count stays right-aligned at every supported sidebar width.
- Updated `SidebarFrame.tsx`, `PluginsSidebar.tsx`, and `LibraryLayout.tsx` so the footer rows share the tree rows' left inset (`px-2.5`) instead of sitting on their own indent.
- Focus outlines and hit areas are unchanged — one `<button>` per row, no nested controls, no overflow-hidden ancestors clipping focus rings.
- Added a changeset and regression tests (`describe.each` over the default and narrowest supported sidebar widths) covering the new class-level layout invariants.

## Acceptance criteria
- [x] The reminder line and the CHANGE REQUESTS row share the left inset of the tree rows above them and a consistent vertical gap.
- [x] The reminder truncates with an ellipsis and a tooltip when the sidebar is narrow; "Finish now" stays a visible link.
- [x] The count on the CHANGE REQUESTS row stays right-aligned at every supported width.
- [x] Focus outlines and hit areas are unchanged.
- [x] A visual regression test at the default and the narrowest sidebar width.

## Verification
- `NODE_ENV=test pnpm typecheck` clean across all 7 projects.
- `NODE_ENV=test pnpm test`: 6846 passed / 1 skipped / 0 failed.
- `pnpm -r run build` green; design-system ratchet 206 vs baseline 215 (no regressions).
- Full local docker boot from source (`deployment/docker-compose.build.yml`): app healthy at `/api/health`, served CSS/JS verified to contain the expected utility classes and DOM structure for both the default and narrowest sidebar widths.

---
## Summary by cubic
Fixes the sidebar footer's alignment so the setup reminder and CHANGE REQUESTS row sit on the same grid as the tree rows above them, addressing the `Sidebar-Footer-Aligned` ticket. Previously each footer row brought its own inset, hairline, and margin—the reminder sat four pixels right of the tree and pressed against the dock—and now `SidebarFrame` owns the footer group with one rule, one gap, and a shared `SIDEBAR_ROW_INSET`.

**Notes**
- Moved the setup reminder out of `PluginsSidebar` into a footer row (`IntegrationsSetupReminder`); `PluginsSidebar` no longer accepts `attentionCount` or `onFinishSetup`.
- Change request rows keep `SIDEBAR_ROW_INSET` and take their caret indent inside the row (`SIDEBAR_CHILD_INDENT`), so the header and its children move together and hover and focus still span the full column.
- At narrow widths the reminder truncates with a tooltip while "Finish now" stays visible, and the CHANGE REQUESTS label truncates so its count stays right-aligned.
- The footer group hides itself when both rows render nothing, so no stray hairline appears at the bottom of an empty footer.
- Adds regression tests at the default and narrowest sidebar widths, and updates `header-band.test.ts` so imports of `HEADER_COLUMN_TOP` aren't mistaken for CSS classes.

<sup>Written for commit 190eecb95b1789f1f96d4a3d022c6c626622eb69. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Bevel-Software/Hexis/pull/277?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


h3xxit:
All three cubic findings accepted and fixed in `e35334cf`.

**1. `IntegrationsSetupReminder.tsx` — the truncation comment was backwards.** Correct: `truncate` cuts the tail, the count leads `said` and survives, `"setup."` is what the ellipsis eats. The tooltip's job is the whole sentence, not rescuing the number. Comment rewritten to say that.

**2. `PullRequestsForMe.tsx` — `pl-[29px]` baked the shared inset into a child's number.** Agreed, and this is the drift the single export exists to prevent. Rather than recompose the same sum (Tailwind scans class names as literals, so `pl-[10+19px]` cannot be built at runtime and a composed constant would still be two numbers agreeing by hand), the indent moved *inside* the row:

- the request row and the error row now carry `SIDEBAR_ROW_INSET`, like every other row in the column;
- their content carries `SIDEBAR_CHILD_INDENT` (`pl-[19px]`, the caret slot plus its gap), exported from `SidebarFrame` beside the inset it is added to.

The header and the rows under it now move together by construction — change `SIDEBAR_ROW_INSET` and both move, because both read it. The 19px is now only the caret geometry and no longer encodes the inset at all. The row still spans the full column, so its hover fill and focus ring are unchanged.

**3. `SidebarFooterGrid.test.tsx` — the `overflow-hidden` ancestor claim was unearned.** Correct: the frame's `<aside>` is `overflow-hidden`, and has to be — that is what lets the column keep its width and slide out intact while the frame animates to zero. It clips the tree rows exactly as much as the footer rows, so it is not what this PR could regress. The test now walks the chain from each row up to `data-sidebar-column` and names any clip it finds there, and the comment says why the column is the ceiling.

**Verification at `e35334cf`:** `NODE_ENV=test pnpm typecheck` clean across all 7 projects (exit 0); `NODE_ENV=test` core-frontend suite 176 files / 2402 tests passed, 0 failed (exit 0); design-system ratchet 206 vs baseline 215, no regressions.

h3xxit:
Accepted and fixed in `190eecb`, plus a merge with `dev` in `273edd2`.

**The finding on `SIDEBAR_CHILD_INDENT`** — correct on every point, and I checked each against the source before changing anything:

- `FileExplorer`'s tree rows carry `w-3.5` (14px) and `gap-1.5` (6px), not 12 and 7.
- Children there are indented `10 + depth * 13` (`indentFor`, `FileExplorer.tsx:94`) — a depth ladder, so a file genuinely does not start under its folder's name.
- The 19px is `PullRequestsForMe`'s own header row: `w-3` + `gap-[7px]`.

So the numbers were right and the attribution was wrong. The comment now names the CHANGE REQUESTS row as the source and says the constant is only correct for children of *that* row — a parent with a different caret needs its own number. The tree analogy is gone, replaced by what the two ends of the column actually do share: the inset, `SIDEBAR_ROW_INSET`.

While there, the old comment's literal `pl-[29px]` is dropped too. Tailwind scans source text rather than parsing it, so quoting that class in prose was enough to emit a `.pl-[29px]` rule no element used.

**Merge with `dev` (`273edd2`)** — `dev` moved on to the header-band work, which touches the other end of this same column. Two textual conflicts plus one that git could not see:

- `SidebarFrame.tsx`: `dev` swapped the column's `pt-4` for `HEADER_COLUMN_TOP`, this branch named that element `data-sidebar-column`. Both kept.
- `SidebarFrameResponsive.test.tsx`: both sides rewrote the same assertion — `dev` wrapped the header in the band, this branch wrapped the footer in a group, so each had deleted the other's wrapper from its picture of the column. The merged assertion states what is true of both ends now: band, contents, footer group, three siblings in one column.
- `header-band.test.ts`: a semantic conflict. `tagsAround` slices from a marker hit to the next `>`, and the **import** of `HEADER_COLUMN_TOP` is a hit with no `<` before it, so the fallback handed the assertions a 4086-character window over the constants below. This branch's `SIDEBAR_FOOTER_SLOT` landed in it and its `pt-2` read as a second offset on the column. Hits that are not inside a tag are skipped now. I confirmed the guard still bites: injecting a real `pt-[34px]` on the column tag still fails the test.

**Verification at `190eecb`:** `pnpm -r run build` exit 0; `NODE_ENV=test pnpm typecheck` exit 0 across all 7 projects; design-system ratchet 206 vs baseline 215, no regressions. Full core-frontend suite result follows once it finishes — the layout and theme files most affected by the merge (`header-band`, `HeaderAlignment`, `SidebarFrameResponsive`, `SidebarFooterGrid`, 9 files / 91 tests) are already green.

h3xxit:
Suite result promised above, at `190eecb`: full core-frontend suite **178 files / 2434 tests passed, 0 failed** (exit 0), with `NODE_ENV=test pnpm typecheck` exit 0 alongside it. Green including the three files the merge with `dev` touched (`header-band`, `HeaderAlignment`, `SidebarFrameResponsive`).

PR #279, opened 2026-09-19
Title-Bar-Aligned-With-Sidebar — The page title bar lines up with the sidebar header

h3xxit:
Ticket: `Data/Engineering/Knowledge/Hexis/Tickets/Title-Bar-Aligned-With-Sidebar.md`

## Summary

The sidebar header row and the page title bar are unified into one shared band so they render at the same height with aligned top edges, across the file/skill/tool/library pages that use the knowledge page header.

- Introduced a shared height token (`--spacing-header: 48px`, `packages/core-frontend/src/shared/theme/tokens.css`) and a shared band module (`packages/core-frontend/src/shared/theme/header.ts`) consumed by both the sidebar header and the page header, so no page keeps a private header height.
- Added tests covering the shared height token and a layout test asserting equal computed heights for the sidebar header and the page header (`packages/core-frontend/src/shared/theme/__tests__/header-band.test.ts`, `packages/core-frontend/src/modules/layout/__tests__/HeaderAlignment.test.tsx`).

## Verification

- Booted the app from source and measured a real Chromium render: sidebar header row and page title bar are both 48px tall with both top edges at 12px, at 1280/1440/1920px and across collapse, reopen, and sidebar resize (min/default/max) — 8 states, all identical. File/library/plugin/skill/tool page headers all measured 48px.
- `pnpm -r build && pnpm typecheck && pnpm test` all green (core-frontend 176 files/2414 tests, core-backend 258 files/4146 tests, including the new tests above).

## Known caveat

AC1's baseline clause is met to within ~3.8px, not exactly: the band centers its contents, and the two sides carry differently sized text (13px pill vs 26px display heading), so centering leaves baselines ~3.8px apart out of 48px total height. Band heights and top edges are exactly equal; the previously reported ~20px misalignment is resolved. Exact baseline agreement would require `items-baseline`, which would knock the pill's chip off-center and misplace the page header's badges/icon buttons. Flagging for reviewer judgment on whether the clause is meant literally.

---
## Summary by cubic
Aligns the page title bar with the sidebar header row so both render as one band at the same height, fixing the ~20px misalignment on Library pages.

Both rows now share a height token (`--spacing-header`) and a top offset, and the band is the first row of every page: breadcrumbs and back links that used to sit above now ride on it. Page titles truncate instead of wrapping; counts, folder paths, and descriptions now sit below the band; the sidebar's header row still renders when empty. Plugin title groups clip their breadcrumb instead of letting it slide over the action buttons, and the back link's negative margin is now a named constant (`HEADER_BAND_LEAD`). Two tests guard the alignment: one reads the sources to confirm no page keeps a private header height, and one compiles the production stylesheet with Tailwind and measures rendered heights.

<sup>Written for commit c832c3330220ffc2ac51233d21e0b533e57cf9f3. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Bevel-Software/Hexis/pull/279?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


h3xxit:
All six findings accepted and fixed in `be31a0ee`. Per-thread:

**P2 — `PluginPage.tsx:299`, the breadcrumb sits above the band.** Correct, and wider than the one file: the personal and locked plugin pages lead with the same breadcrumb, and the skill and tool pages lead with a back link plus `mt-4`. Five pages whose title bar therefore never shared the sidebar's top edge, which is the ticket's whole point — the band's top edge is the seam, not just its height.

That navigation now rides **on** the band as its leading item rather than in a row above it. `PluginBreadcrumb` stops short of the current page while it is up there (the `<h1>` beside it *is* the current page; a crumb repeating the heading it sits next to says the name twice on one line), which also makes the trail a fixed width so it never competes with the title. `LockedPluginView`'s hand-copied trail is retired in favour of the shared component. The error and not-found returns on the skill and tool pages still lead with the back link on its own row — they have no title bar to hold a line with. `LibraryRoutes.test.tsx` now asserts the breadcrumb and the `<h1>` are both inside `PAGE_HEADER_TESTID`, so the row cannot come back.

**P2 — `KbPageHeader.tsx:228`, overflow at narrow widths.** Valid: `flex-wrap` used to let the action cluster drop to a second line, and a band cannot wrap. Rather than add a narrow-screen overflow menu (a product change this ticket did not ask for), the row is now explicit about what gives first: the title and its status chips share one `min-w-0 flex-1 overflow-hidden` group and the action cluster sits outside it as `flex-none`. The row gives up title text — still fully readable via the `title` attribute — and then clips chips, and the Share/Edit controls stay reachable at any width instead of being pushed off the page. If a real overflow menu is wanted, it is worth its own ticket; this makes the failure mode bounded in the meantime.

**P2 — `HeaderAlignment.test.tsx:64`, the test supplied the rule it measured.** Fixed properly: the test now hands `src/index.css` — the application's real entry, Tailwind and the tokens and the typography plugin — to Tailwind's own compiler and injects the CSS that comes out. The band's classes are read off `HEADER_BAND` so the compiled candidate set cannot drift from the rendered one, and nothing about the height is restated in the test any more: not the token's value, not the `.h-header` rule. Verified by mutation — renaming `--spacing-header` in `tokens.css` now fails the suite where before it would have passed. One caveat documented in the file: `@layer` is unwrapped before injection, because happy-dom does not implement cascade layers; with a single stylesheet in the document there is nothing for layers to order, and only braces move, never declarations.

**P3 — `SidebarFrame.tsx:283`, comment says 16px.** Right, and the same comment then said 12px two lines later. Rewritten to say the prototype's 16px is the number that did *not* survive and `HEADER_COLUMN_TOP` replaces it with 12px.

**P3 — `tokens.css:113`, the arithmetic does not add up.** Right, and your 31.5px is the correct pill height. The claim was wrong, not the value: 48px is not a sum, it is the smallest 4px step that holds both sides at their present size. The comment now says that, with both measurements — the 31.5px pill and the 29.9px `--text-display` heading — and the 9px of air the band leaves around the taller one.

**P3 — `header-band.test.ts:108`, the enumeration can go stale.** Fixed with exactly the failsafe you describe, in both directions: `header-band.test.ts` now walks every non-test `.tsx` under `src/` and asserts the discovered set of files carrying a header-row marker equals `HEADER_ROWS`, and likewise for `HEADER_COLUMN_TOP` and `HEADER_COLUMNS`. A page added later that keeps a private header height fails the suite; so does a listed file that stopped rendering a row.

**P3 — `header-band.test.ts:129`, the `pt-[34px]` ban is file-scoped.** Right, and inconsistent with the file's own stated rule. The row and column tag extraction is now one helper, and the column check is scoped to the opening tag that carries `HEADER_COLUMN_TOP`. It is also generalised while it is there: any literal `pt-` class on that tag fails, not just the one historical value, since a second number there re-creates the seam whatever it says.

Local verification on the new head: `core-frontend` typecheck clean, 176 files / 2419 tests green (up from 2414 — three new cases), eslint clean on every touched file.

h3xxit:
Correction to the test count in the comment above: the suite on `be31a0ee` is **176 files / 2417 tests**, up 3 from 2414 (two staleness failsafes in `header-band.test.ts`, one breadcrumb-on-the-band assertion in `LibraryRoutes.test.tsx`). Everything else in that comment stands — typecheck clean, eslint clean on the touched files.

h3xxit:
Re-review of `be31a0ee` raised two new findings (review `5255676848`). Both accepted and fixed in `ba80b191`.

**P2 — `PluginPage.tsx:301`, breadcrumb can overlap the action cluster at narrow widths.** Correct, and it is a defect I introduced in `be31a0ee`: the trail is `flex-none` inside a group that was allowed to shrink. Flex shrinks an item's *box*, not its contents, so once the group narrowed past the trail's width the trail rendered outside its own box and across the manifest and action buttons — a control you cannot click because a breadcrumb is sitting on it.

The group now takes the row's shortfall deliberately (`flex-1 min-w-0`) and clips it (`overflow-hidden`): the title truncates first, the trail goes last, and the actions keep their width and their hit areas. That is the first of the two options you offered, chosen over hiding the trail at a breakpoint so that all four page headers now degrade the same way — it is the same treatment `KbPageHeader`'s title-and-chips group got for the previous P2. `PersonalPluginPage` has the identical structure (same breadcrumb, same `justify-between`, same `flex-none` `PageActions`) and got the identical fix.

Two bands are deliberately left alone, and I would rather say why than change them silently:

- **`SkillPage` / `ToolPage`** have no shrinkable group. Their children are the back link (`flex-none`), the heading (`min-w-0 truncate`, which shrinks its content properly), and `flex-none` badges/actions. A too-narrow row overflows the band's right edge; it does not put one control on top of another, because nothing there shrinks its box past its contents. Wrapping them in a clipping group would also fight the `-ml-2` below — `overflow-hidden` would clip the 8px the lead deliberately hangs outward.
- **`LockedPluginView`** anchors nothing to the right (no `justify-between`, no `ml-auto`), so there is no fixed cluster for the trail to collide with.

If you read the overflow-without-overlap case on those three as in scope too, say so and I will treat it as a separate pass — I did not want to restructure two more headers on an inference.

**P3 — `ToolPage.tsx:155`, the undocumented `-ml-2`.** Right on both counts: undocumented, and coupled to a value in another file. It is now `HEADER_BAND_LEAD` in `shared/theme/header.ts`, used by both the skill and tool pages, with the coupling written down — `Button`'s quiet `sm` size carries `px-2` (`QUIET_SIZE.sm`), and cancelling exactly that is what puts the link's *text* on the same column edge every other page's title starts on, while the button keeps the hover background its padding exists for. The docstring names `QUIET_SIZE` so the dependency is greppable from `Button.tsx`, which is the direction it would be broken from.

The other seven threads are from the first review (`5255486551`) and were fixed in `be31a0ee`; three of them show a `be31a0ee` commit id only because GitHub re-anchored their line positions, not because they were re-raised — the re-review did not repeat any of them.

Verification on `ba80b191`: `core-frontend` typecheck clean, 176 files / 2417 tests green, eslint clean on the touched files.

h3xxit:
Re-review of `598f508e` raised one finding (review `5255723705`), on the changeset. Accepted and fixed in `c832c333`.

**P3 — the tangled sentence in `.changeset/title-bar-aligned-with-sidebar.md`.** Right on all three counts, and thank you for separating the prose from the behaviour — "says what" had no complement, the colon restarted the sentence, and `are clipped` read back onto the singular `the title`. A changelog entry is published prose and that one had to be parsed twice.

Rewritten close to your suggestion, with the subject carrying its complement and the mechanism after the colon agreeing with itself:

> A band that cannot wrap has to give something up when a narrow window runs it out of room, and every title bar now decides that the same way: the title truncates first, then whatever shares its group — a plugin page's breadcrumb, a file page's status chips — is clipped, while the controls beside them keep their width and stay clickable. That is a Share button still reachable on the file page, and a breadcrumb that never lands on a plugin page's manifest and action buttons.

I kept the second sentence rather than dropping it as your suggestion did, because the two concrete consequences are the part a reader of a changelog actually acts on. The behaviour described is unchanged — no code moved in this commit.

That clears every thread on this PR. For the record, across three rounds: seven findings from review `5255486551` fixed in `be31a0ee`, two from `5255676848` fixed in `ba80b191`, one from `5255723705` fixed here. The only thing I pushed back on was scope, on the thread above — `SkillPage`, `ToolPage` and `LockedPluginView` were left unrestructured because none of them has a shrinkable group wrapping a fixed-width child, and a Chromium check at 320px confirmed the tool page truncates cleanly with nothing anchored right to collide with.

Verification on `c832c333`: documentation only, no code change, `changeset status` parses the entry. The last full run, on `598f508e`, was typecheck clean in every package with core-frontend 176 files / 2417 tests and core-backend 258 files / 4146 tests green.

PR #280, opened 2026-09-19
Tool-List-Survives-Bad-Manual — One malformed .tool file no longer hides every other tool

h3xxit:
Ticket: Data/Engineering/Knowledge/Hexis/Tickets/Tool-List-Survives-Bad-Manual.md

## Summary
- A malformed `.tool` manual was already isolated per-file by `scanDisk` (a YAML error, schema violation, or reserved-variable reference did not hide other tools), but the failure was silent: nothing logged it, named the file, or surfaced it in any response, so a vanished tool looked identical to an outage.
- Adds an `invalid: [{ path, reason }]` field (with no credential values) to `list_tool_setup`, `GET /api/tools`, and `GET /api/secrets/tools`, naming the bad file and the validation location (line/field where available).
- A single `IToolManualService.listAccessibleCatalog(userEmail) → { tools, invalid }` (one scan, one access pass, single-flight) and an exported `describeManualFault(err)` back the new field; fixing or disabling the bad file restores it on the next listing with no reconnect. (An earlier `listInvalid` was removed in review: two calls meant two scans and two snapshots.)
- AC4 (Library page) is satisfied server-side: `GET /api/secrets/tools` carries `invalid`; no UI rendering was added since a validating manual editor is out of scope per the ticket.

## Changed files
`packages/core-backend/src/modules/tool-manuals/{tool-manuals.service.ts,tool-manuals.contract.ts,tool-manuals.tools.ts,tool-manuals.routes.ts}`, `packages/core-backend/src/modules/secrets-vault/secrets-vault.routes.ts`, new tests `tool-manuals.invalid.test.ts` and `tool-manuals.tools.test.ts`, `.changeset/tool-list-survives-a-bad-manual.md`.

## Testing
- `pnpm typecheck` and `pnpm build` green across all packages.
- core-backend vitest: 260 files, 4195 passed, 1 skipped (includes direct route tests for `GET /api/tools` and `GET /api/secrets/tools`).
- Local from-source compose boot: verified all 5 ACs end-to-end (multiple malformed manuals coexisting with valid tools, `invalid` reasons pinpointing yaml/schema/reserved-variable/credential faults with zero credential leakage, recovery on the same long-lived MCP session with no reconnect, and catalog registration + Library route isolation).

---
## Summary by cubic
Reports malformed `.tool` files instead of silently skipping them, so a missing tool no longer looks like an outage. Every listing that carries the catalog now carries `invalid: [{ path, reason }]`: `list_tool_setup`, `GET /api/tools`, and `GET /api/secrets/tools`.

**Bug Fixes**
- One file's fault is isolated: valid tools stay listed and callable, and fixing or deleting the bad file restores it on the next listing with no reconnect.
- `reason` locates each fault by field name or ordinal and never echoes the file's content; refusals are access-gated like catalog entries.
- Namespace collisions and unreadable files report through the same channel; warnings log once per change.
- `tools` and `invalid` come from one single-flight scan, so every surface describes the same snapshot; both HTTP routes and the MCP tool now have end-to-end route tests.

<sup>Written for commit e1935ecb4a8022df70ccfb3a20ad55c964ed6b1a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Bevel-Software/Hexis/pull/280?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


h3xxit:
Review addressed — all seven findings accepted, none disputed. Summary for the human reviewer; cubic re-reviews the head on its own.

**P1 — a rejected value echoed into `reason` (contract.ts:285).** Agreed, and the fix is upstream rather than in `describeManualFault`: `redactSecret` knows this process's tokens and the shape of a URL, and no scrub can recognise an arbitrary author-written string as a credential, so a message that interpolates one cannot be made safe after the fact. The normalizer's refusals now name the FIELD and the RULE instead — ``tool `id` must be lowercase snake_case``, ``unknown `.tool` `type` (expected http, mcp, inline, or absent)``, ``variable "X" has an invalid `scope` `` — and a `variables` entry whose name has not yet passed its own `[A-Za-z0-9_]+` check is located by index (`` `variables[1].name` ``) rather than quoted. New table-driven tests plant the same literal token in `id`, `type`, a variable name and a variable scope and assert it reaches no `reason`.

One deliberate exception, stated so it can be argued with: `assertStringHeaders` still writes `` `headers.Authorization` must be a string ``. The rejected VALUE is not echoed there — the key is the field locator, the same role `variables[i]` plays — and the guarantee the doc now records is "names the field, never the value".

**P2 ×4 — two scans per cold listing, two snapshots in one response (routes.ts:346, secrets-vault.routes.ts:208, tools.ts:201, contract.ts:316).** One root cause, fixed at the service rather than at each call site: `listAccessible` + `listInvalid` is replaced by a single `listAccessibleCatalog(email) → { tools, invalid }` that runs one scan and one `canReadBatch`. `listInvalid` is gone from the contract entirely — nothing needed the refused half alone, and leaving it there leaves the pairing available to get wrong again. Underneath, `scan()` is now single-flight (`inFlightScan`), so any reader arriving during a walk shares it; `invalidate()` clears that promise as well as the cached value, so a caller arriving after a merge is not handed the scan that started on the tree the merge replaced — the rule `TtlCache`'s generation token already kept for the cached value. Two tests: three concurrent listings ⇒ 1 walk, 3 ACL passes (per-caller, correctly not shared) and identical snapshots; and an `invalidate()` mid-scan ⇒ 2 walks, with the second listing seeing the newly added file.

**P2 — `printable` on the refusal warning (service.ts:585).** Agreed; a KB filename may legally contain a newline. Both path and reason now go through `printable` (which quotes what it escapes, so the literal quotes came out of the template). Test: a `.tool` whose filename embeds `\n[tool-manuals] catalog is empty` produces exactly one warning line, with the newline escaped.

**P3 — false pass as root (invalid.test.ts:263).** Agreed; `test.skipIf(process.getuid?.() === 0)` now, so a root container reports the coverage it did not get.

core-backend: 259 files, 4177 passed / 1 skipped. `pnpm typecheck` green across all packages. eslint clean on the changed files (the one error under `packages/core-backend/src/modules/tool-manuals/` is the pre-existing `no-explicit-any` at `tool-manuals.mcp-oauth.test.ts:96`, which reproduces on the untouched tree). Branch synced with `origin/dev` first — merge, no conflicts.

h3xxit:
Second review pass addressed — both findings on `029dee5e` accepted.

**P1 (`tool-manuals.service.ts:1320`) — an invalid `scope` still copied `name` into `reason`.** Correct, and it shows the rule I drew last time was in the wrong place. I had kept a variable's `name` in the messages *downstream* of its `[A-Za-z0-9_]+` check, arguing a name is a locator rather than a value — including in the reply above, where I drew the same line around header keys. That argument does not survive contact with `sk_live_51H8ffGG`, which passes `[A-Za-z0-9_]+` and is a credential. Passing a grammar test makes a string legal, not repeatable.

So rather than patch the one message, the guarantee is widened to the one that can actually be checked and is now written at the top of `describeManualFault`:

> **A reason repeats nothing the file said.**

Every remaining interpolation goes with it — not just the scope refusal cubic anchored on:

| was | now |
|---|---|
| `variable "X" has an invalid `scope`` | `` `variables[0].scope` is invalid (expected admin\|user)`` |
| `variable "X" oauth must be an object`, `duplicate variable "X"`, and the rest of the oauth block | `` `variables[0].oauth…` ``, `` `variables[1].name` duplicates an earlier entry's `` |
| `` `headers.Authorization` must be a string `` | `` `headers` entry 3 of 4 must have a string value `` |
| `` `.tool` "X" references the reserved variable … `` | `` this `.tool` references the reserved variable … `` |
| `tool namespace "X" is reserved…`, `` `.tool` "X" declares `remote: true`… `` | the same sentences without the name |
| `manual "X" resolves to the secret-variable namespace "Y"…` | the same without either — `Y` is `X` transformed, which is no better |
| SSRF labels `` `.tool` "X" url `` / `` "X" healthCheck.url `` | `` `url` `` / `` `healthCheck.url` `` |

What stays is text matched against a FIXED list of ours — a reserved variable name, a reserved namespace — where naming the one that was hit repeats nothing the file taught us. Nothing was lost in diagnostics: the name was always redundant with `path`, which is in the same `invalid[]` entry.

Seven new cases cover it, each planting an identifier-legal token (`sk_live_…`, underscores) in the field that used to echo it: the bad-scope case cubic named, a malformed `oauth`, a duplicate declaration, a manual name on a reserved-variable reference, a manual name on a `cli`-plus-`remote: true` refusal, a header key with a non-string value, and a namespace collision between two token-named files.

**P3 (`tool-manuals.invalid.test.ts:477`) — the test slept through the poisoning case.** Also correct: `await first` came after the assertions, so a stale scan overwriting the fresh cache entry on settle — the exact failure `TtlCache`'s generation token exists to prevent — would not have failed anything. The test now settles the stale scan first, then reads a third time with no intervening `invalidate()` and no clock advance: `walks` must stay 2 (a third walk would mean the cache was emptied) and the post-merge catalog must still be what comes back (the pre-merge one would mean it was overwritten).

core-backend: 259 files, 4184 passed / 1 skipped (up from 4177). `pnpm typecheck` green across all packages. eslint clean on the changed files apart from the pre-existing `no-explicit-any` at `tool-manuals.mcp-oauth.test.ts:96`. Branch re-synced with `origin/dev` first — already up to date.

h3xxit:
Architecture review addressed — both findings accepted. They were right: `list_tool_setup` had a direct test and the service had plenty, but `GET /api/tools` and `GET /api/secrets/tools` were covered only *through* the service. A field that exists in the service and never reaches the wire is not a feature anyone can use, and nothing in the diff ruled that out.

**New: `packages/core-backend/src/modules/tool-manuals/__tests__/tool-manuals.invalid.routes.test.ts`** — both routers mounted on one express app over ONE real `ToolManualService`, reading a real temp knowledge base containing a real malformed manual. Deliberately not stubbed: a stubbed service cannot tell us whether the route serializes `invalid` at all, which is precisely the claim under review.

Ten cases:

| | `GET /api/tools` | `GET /api/secrets/tools` |
|---|---|---|
| valid tools all listed, refused file named with a located reason | ✓ | ✓ |
| refusal withheld from a caller who may not read that path | ✓ | ✓ |
| unauthenticated call refused | ✓ | ✓ |
| recovery against the same running server — file fixed, and file deleted | ✓ | — |
| `?path=` narrows `invalid` as it narrows the manuals, both ways | — | ✓ |

Plus one that both surfaces report the same refusal from the same snapshot.

**The tests were checked for bite, not just for green.** Dropping `invalid` from either route handler (`res.json({ tools })`) fails 7 of the 10. The three survivors are structural and worth naming: the two 401 cases never reach the body, and the cross-surface comparison passes because both sides break together.

**Two stale claims in the PR description above, which I cannot edit from here** — flagging them so they are not read as current:
- "A new `IToolManualService.listInvalid(userEmail)`" — `listInvalid` was **removed** in `029dee5e` in favour of a single `listAccessibleCatalog(email) → { tools, invalid }`, after cubic pointed out that fetching the two halves separately meant two scans and two possible snapshots.
- "core-backend vitest: 259 files, 4169 passed" — now 260 files, 4195 passed / 1 skipped.

The rest of the description still holds. `pnpm typecheck` green across all packages; the new file lints clean (the 9 eslint errors under these two modules are pre-existing `no-explicit-any` in `db-secrets-vault.oauth.test.ts` and `tool-manuals.mcp-oauth.test.ts`, both reproduced on the untouched tree). Branch re-synced with `origin/dev` first — merged `ac5f8a49`, frontend-only, no conflicts.

No new cubic findings on `bffd40f5`; the round-2 threads are now marked outdated.

PR #281, opened 2026-09-19
Title-Bar-Aligned-With-Sidebar — The page title bar lines up with the sidebar header

h3xxit:
Ticket: `Data/Engineering/Knowledge/Hexis/Tickets/Title-Bar-Aligned-With-Sidebar.md`

Follow-up to #279 (merged), which fixed a regression found on staging.

## Summary

Staging testing on the merged #279 found that the file page's title bar opened 54px below the sidebar header row at 1280/1440/1920 — a structural regression, unaffected by collapse/resize. Root cause: `FileViewer` rendered `<EditorTabs />` above `<KbPageHeader>` inside the same document column, so the 48px band was correct but the tab strip (36px + 18px margin) pushed it down.

- `KbDocumentShell` now takes the title bar as a `header` slot and renders it as the column's first row in all three variants (prose, rail, full-bleed), on the shared `HEADER_COLUMN_TOP` offset — the mirror of how `SidebarFrame` owns the header row on the sidebar side of the seam. The tab strip stays in the same column, below the title. Full-bleed pages now get the shared offset for the first time.
- Closed the test gap that let this regression through: `HeaderAlignment.test.tsx` previously rendered `KbPageHeader` standalone (so `EditorTabs` never existed in the test document) and only asserted equal heights, never equal top edges. It now renders the real page through `KbDocumentShell`/`FileViewer` and asserts top-edge equality too. Verified by mutation: reverting the first-row contract fails 4 tests across 3 files.

## Verification

- Booted from source; real Chromium measurement through the real `KbDocumentShell` (with a real tab strip) for prose-with-tabs, prose-no-tabs, rail and full-bleed: all report 0px top-edge and height delta at 1280/1440/1920 and across collapse, reopen and a min→default→max sidebar drag. A negative control reproducing the pre-fix shape reproduces staging's failure, confirming the test actually catches this class of bug.
- `pnpm -r build && pnpm typecheck && pnpm test` green: core-frontend 176 files/2426 tests, core-backend 258 files/4146 tests (1 skipped).

## Known caveats (carried from #279)

- AC1's baseline clause is met to within ~3.2–3.8px, not exactly, due to centring differently-sized text in one fixed-height band; exact agreement would require `items-baseline`, which would misplace other elements sharing the row.
- `LockedPluginView` has not been rendered by any harness, local or staging, and remains unverified.

---
## Summary by cubic
Fixes the file page title bar so it lines up with the sidebar header row. It previously opened 54px below the sidebar row in the standard layout and 12px above it in full-bleed; it now starts on the same shared offset in both.

**Changes**
- `KbDocumentShell` takes the title bar as a `header` slot and renders it as the first row in every layout variant.
- Tests now require the header to share the column with the page content, so a regression that hoists it above the column fails locally.
- The tab strip stays below the title, so it no longer pushes the title bar down.
- Alignment tests render through the real shell, so this regression would be caught locally.

**Known caveats**
- The baseline alignment target is met within ~3.2–3.8px because differently sized text is centered in a fixed-height band.
- `LockedPluginView` remains unverified by any harness.

<sup>Written for commit 0c3888895352c7b60dd6e9a07605ddcbd00aa4e4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Bevel-Software/Hexis/pull/281?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


h3xxit:
One finding on this head (review `5255963080`). Accepted and fixed in `0c388889`.

**P2 — `KbDocumentShell.test.tsx:35`, the assertion does not tie the band to the column.** Correct, and it is the sharper version of the mistake this whole PR exists to fix. `previousElementSibling === null` says the band is first among *its* siblings; it says nothing about whose siblings those are. A regression that hoisted the header into a wrapper of its own — above the column, or above the shell box — satisfied it exactly as well as the correct shape did. That is also the regression that actually reaches the seam, because a wrapper outside the column does not carry `HEADER_COLUMN_TOP` and anything at all can be rendered underneath it. The docstring claiming the test asserted "the band opens the column" was writing a cheque the assertion did not cover, which is close to how the file page shipped broken in the first place.

All three variants now assert what you describe, plus the containment:

```ts
expect(band.parentElement).toBe(content.parentElement);
expect(band.previousElementSibling).toBeNull();
expect(band.nextElementSibling).toBe(content);
expect(screen.getByTestId('kb-document-shell')).toContainElement(band);
```

Mutation-checked rather than asserted, since a strengthened test that cannot fail is the same bug again. Hoisting the prose header into its own wrapper above the column:

```
× puts the header first in the column (prose)
  → expected <div …><h1></h1></div> to be <div …><div></div></div>
```

The old assertion passed that shape; the new one fails on `parentElement`. Source restored, 9/9 green.

Verification on `0c388889`: `core-frontend` typecheck clean, 176 files / 2426 tests green, eslint clean. Test-only change — no source moved, and the changeset stays accurate as written.

PR #282, opened 2026-09-19
Tool-Names-Stay-Readable — Card and row names keep a readable width and badges wrap under them when space is short

h3xxit:
Ticket: `Data/Engineering/Knowledge/Hexis/Tickets/Tool-Names-Stay-Readable.md`

## Summary
- A card's name now keeps a floor of 12 characters while badges sit beside it; the badges wrap to a second line instead of taking space from the name.
- The full name stays in `title` and in the DOM (for screen readers) even when visually truncated.
- Plugin index rows and the tool page header follow the same `NameWithBadges` component and rule. The plugin row gains a `title` it previously lacked.
- `ItemMenuFrame`'s grid track is floored at `minmax(0,1fr)` (was implicit `auto`), fixing a pre-existing overflow where a long plugin name laid the row out wider than its column and left badges no room to wrap.
- The tool page header takes truncation + `title` but no floor, since it carries no badges and a floor there would just push the row wider than its band.

## Test plan
- `pnpm test` — 258 files, 4146 passed / 1 skipped / 0 failed.
- `tsc --noEmit` on core-frontend — clean.
- Tests cover a 60-character name on a card, a row, and the page header; a missing display name (no empty tooltip); two tools of the same name in different plugins (stay separately addressable).
- Pixel measurements taken in chromium against the production Tailwind stylesheet: card/row badges wrap under the name at narrow widths and share a line from ~339px/324px on; tool page header band stays at 48px with 0 overflow at phone widths, matching the pre-ticket baseline.

---
## Summary by cubic
Fixes unreadable tool and plugin names on narrow cards and rows by giving the name a 12-character floor and letting badges wrap below it when there isn't room for both.

- Library cards, plugin index rows, and the tool page header now share the same `NameWithBadges` layout.
- Truncated names keep their full text in the DOM and in `title`, so screen readers and hover can still surface the whole name.
- Plugin rows gain a `title` they previously lacked, and their grid track now lets the row shrink to its column instead of overflowing it.
- The tool page header truncates with a `title` but takes no floor and cannot wrap, because its one-row band has no second line to use.
- Rows with badges can now be two lines tall on narrow windows; which badges appear is untouched.

Ticket: `Data/Engineering/Knowledge/Hexis/Tickets/Tool-Names-Stay-Readable.md`

<sup>Written for commit 404be51a17013ca536e98c4e1a3a7207a3784389. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Bevel-Software/Hexis/pull/282?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


h3xxit:
Review addressed at `404be51a`.

**cubic's one finding** (`name-with-badges.test.tsx:121`, P3) — accepted and fixed. The trailing space in `not.toContain('flex-wrap ')` read like it was disambiguating `flex-wrap` from `flex-nowrap`, but `flex-wrap` is not a substring of `flex-nowrap`, so it disambiguated nothing and would have let the assertion pass the day `flex-wrap` landed as the final token in the class list. Now asserts the bare class, with a comment saying why the space was wrong so it does not come back.

**Merged `origin/dev` in** (`0482932c`), which it needed — one conflict, in `LibraryCard`'s name row. dev removed the lifecycle badge from the card outright (`1c6d49cb`, "a skill's lifecycle and owner are no longer the platform's business") while this branch was moving that same row's badges out of a JSX fragment into a countable array — `NameWithBadges` can only grant the name its floor when it knows whether anything is actually beside it, and five falsy branches in a fragment look exactly like five pills. Resolved by keeping this branch's `NameWithBadges` row and taking dev's removal, which is what dev's own test now asserts ("renders no lifecycle badge, even when handed the removed prop"). Four badge conditions remain, in their old order. The changeset's list of the pills that crowded out the name was updated to match — it named a `Deprecated` the card no longer has.

Green on the merged tree: `pnpm typecheck` 0, `pnpm test` 0 — 483 files, 4404 passed / 1 skipped / 0 failed across all four packages.

PR #284, opened 2026-09-19
unknown-branch-is-404 — A branch that never existed answers 404, and 410 is kept for a branch that was deleted

h3xxit:
Ticket: `Data/Engineering/Knowledge/Hexis/Tickets/Unknown-Branch-Is-404.md`

## Summary

Naming a branch that never existed now answers 404 (`branch-not-found`, "There is no branch named <name>."), while a branch that was cloned or listed and whose remote ref is now gone still answers 410 (`remote-branch-gone`), as before.

The distinction is made from what the platform knows: an on-disk/registered clone, or a branch seen in a prior listing, counts as known and therefore "deleted" (410) rather than unknown (404). Listed-branch memory accumulates across listings rather than being replaced, since the listing that reveals a branch is gone is precisely the one that no longer names it — a restart forgets a deleted branch whose clone was already swept off disk, and it then correctly reads as unknown.

- Frontend keeps the existing deleted-branch screen for 410 and now shows the file-missing-style "Branch not found" screen for 404.
- The agent tool surface carries the same `branch-not-found` / `remote-branch-gone` kinds as the HTTP routes.
- An unreachable host or refused credential is untouched — it remains our failure, not the branch's.

## Test plan

- [x] Unit/integration tests cover unknown, deleted, existing, and inaccessible branches over both the tools and the routes (`pnpm test`: 4163 passed).
- [x] `pnpm typecheck` clean.
- [x] Verified end-to-end against a source build in Docker: unknown branch -> 404 `branch-not-found`; a branch listed then deleted on origin -> 410 `remote-branch-gone`; existing branch -> 200; bad credentials -> 500 with no kind.

---
## Summary by cubic
Previously, a branch name that never existed got the same 410 "no longer exists on the remote" as a branch that was deleted, implying it had once existed. Now the platform returns 404 `branch-not-found` for a name it has never cloned or listed, and keeps 410 `remote-branch-gone` for a branch it had cloned or seen in a listing when origin no longer has it.

- Branch names accumulate in memory and are never evicted: a listed name stays known after it drops out of listings, and a cloned branch's name outlives its workspace after a sweep. The memory is not persisted, so a restart forgets a deleted branch with no clone left on disk, and that name reads as unknown.
- The file page keeps the deleted-branch screen for 410 and adds a "Branch not found" screen for 404.
- Tool responses now carry the same `branch-not-found` / `remote-branch-gone` kinds as the HTTP routes, so callers can switch on the kind instead of the message.
- Unreachable hosts, refused credentials, and unreadable clone probes remain 500s: the platform's failure, not the branch's.

<sup>Written for commit 8f169bb7e96dcf8ec22200516414e0593c0e6fad. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Bevel-Software/Hexis/pull/284?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


h3xxit:
All six review findings are addressed in `8f169bb7`. Every one of the P2s was the same failure in a different place — the record that we once knew a branch getting dropped, which turns a 410 into the 404 this branch exists to stop mis-stating — so they are fixed together.

**P2, eviction at `MAX_LISTED_BRANCHES` (workspace.service.ts:326)** — accepted, and the policy was worse than the cap: a deleted branch stops appearing in listings, so it stops being refreshed, so the LRU throws away exactly the entries the 410 depends on. The cap and the eviction are gone; nothing is ever removed from the set. Growth is one short branch-name string per distinct name origin has ever shown the process (kilobytes for a real repository, reset by a restart), which the spec's "no new persistence" decision leaves as the right trade rather than a persistent store.

**P2, clone-known marker lost on cleanup (workspace.service.ts:341)** — accepted, real: a branch opened by a direct link and never listed had only `branchDirs` as evidence, and `deleteWorkspace` / the orphan sweep drops that. `branchDirs` now has a single writer, `registerBranchDir`, which also records the name in the never-cleared set, so the name outlives the clone. New test: cloned by link with no listing at all, deleted on the host, workspace swept → still 410.

**P2, `catch` swallowing probe failures (workspace.service.ts:346)** — accepted. `isAbsence(err)` is the only "no clone here"; anything else rethrows. It rethrows into the clone failure path, so the call site now logs the probe fault and answers 500 with the clone's own message — we cannot tell which story it is, and guessing either one states something about the branch we do not know. New test drives an `EACCES` probe and asserts the result is neither `BranchNotFoundError` nor any status-carrying domain error.

**P3, missing restart→404 test** — accepted, added: listed, cloned, deleted on the host, workspace swept, then a fresh `service()` (the restarted process) → `BranchNotFoundError` / 404 with the exact message. That is the half that exercises `hasHeardOfBranch` returning false through all three checks.

**P3, keep-alive teardown** — accepted, applied to `workspace.routes.branch-errors.test.ts` and to the sibling `workspace.tools.branch-errors.test.ts`, which had the same pooled-socket wait. Both files now tear down immediately (the four suites together run in ~4.6s).

**P3, changelog overstating the memory** — accepted, but resolved the other way round: rather than documenting the 5,000-name bound, the bound is gone, so the changeset now says nothing is evicted and that a clone's name outlives the clone. The changelog and the implementation agree again.

Full suite green on the new head: 261 files, 4166 passed / 1 skipped; `pnpm build` and `pnpm typecheck` exit 0. `pnpm lint` fails repo-wide on pre-existing frontend `react-refresh` errors — none in any file this PR touches.

PR #286, opened 2026-09-19
Unreadable-File-Says-Access — A change request file the viewer may not read says so, instead of a generic could-not-read line

h3xxit:
Ticket: Data/Engineering/Knowledge/Hexis/Tickets/Unreadable-File-Says-Access.md

## Summary
When a change-request dialog fails to read a file's copy because of a permission denial (403), it now shows a specific message naming the nearest folder with an access rule for that path (falling back to "this file" when it can't be named), instead of a generic could-not-read line. Any other read failure shows a distinct message with the reason and a Retry link that re-fetches the file. The file row in the tree also carries a small lock glyph when the copy is permission-denied, using the same sentence as its tooltip. The base-copy variant shares the same two message forms. The word "honest" no longer appears in either message.

## Testing
- `NODE_ENV=test pnpm --filter @bevel-software/platform-core-frontend test` — 182 files / 2465 tests passed, including the new 403 form, other-error form, retry, and lock-glyph coverage.
- Typecheck passes for core-frontend and core-backend.
- Verified live against a full local boot (docker-compose.build.yml) with a real KnowledgeBase access rule: 403 read produced the access sentence naming the ancestor folder with an access.md rule and the lock glyph/tooltip on the row (no Retry link); an injected 500 produced the other-error sentence with a working Retry that re-fetched and recovered.

---
## Summary by cubic
Fixes the change-request dialog so a file the viewer can't read no longer looks like an outage. Permission denials now show an access message naming a folder to ask an owner of (falling back to “this file”); other read failures show the reason with a Retry link, and no Retry appears for denials.

**Behavior**
- Prefers the nearest access rule whether it grants or denies, so denials point to the folder actually holding the reader out.
- Adds a lock glyph to file rows whose copy was denied, using the same access sentence as its tooltip.
- Applies the same two message forms when the unreadable copy is the base copy.
- Re-reads the fork-point copy when the target branch changes, so the named folder is resolved against the right branch.
- Removes the word “honest” from both messages.

<sup>Written for commit d5f9931d445387156f43953213fb62b4ced4befe. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Bevel-Software/Hexis/pull/286?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


h3xxit:
All four review findings accepted and fixed in d5f9931d (the branch is also now merged up with `origin/dev`).

**P1 — `nearestRuleFolder` infers the folder from grant-only `sources`** ([r4054177215](https://github.com/Bevel-Software/Hexis/pull/286#discussion_r4054177215)) — valid. A 403 is as often a `deny` written nearer the file as it is a missing grant, and the deny is normally the *closer* rule, so the sentence could name the folder that let the reader in — or, where the deny was the only ancestor rule at all, fall back to "this file" with the governing folder sitting right there in the response. `nearestRuleFolder(sources, denials)` now reads the `ancestor` entries of both maps and still takes the nearest; `governingRuleFolder` passes `access.denials`. Three unit tests pin it: deny nearer than the grant wins, a deny-only ancestor is named rather than falling back to the file, and a deny written farther out leaves the nearest grant in place.

**P2 — a base/fork denial leaves the tree row unlocked** ([r4054177220](https://github.com/Bevel-Software/Hexis/pull/286#discussion_r4054177220)) — valid. Rather than assembling `deniedNote` from a second source, the pane's failure expression is hoisted into one `selectedFailure` value that both the pane and the row read, so the two cannot disagree about which sentence a file gets or whether it gets one. New dialog tests: a refused base copy locks the row with the pane's own sentence, and a base copy that merely broke leaves it unlocked.

**P2 — `targetBranch` missing from the fork-point read key** ([r4054177223](https://github.com/Bevel-Software/Hexis/pull/286#discussion_r4054177223)) — valid, and the suggested key is what landed. The `asked` guard is keyed on the same string, so the effect's dependency on `targetBranch` could never take effect. New `useForkPointFile.test.tsx` covers it (it fails on the old key) plus the ask-once-per-branch case it must not break.

**P3 — orphaned JSDoc in `useFileOnBranch.test.tsx`** ([r4054177233](https://github.com/Bevel-Software/Hexis/pull/286#discussion_r4054177233)) — valid. `FAILED_404` and its own doc now sit above the intro comment, which is back directly on its `describe`.

Verified on the new head: `pnpm typecheck` clean across every package, `pnpm build`, `pnpm run notices:check` and `pnpm ds:check` green, core-frontend 188 files / 2589 tests passing. Each new test was confirmed to fail against the pre-fix code.

PR #287, opened 2026-09-19
Upload-Feedback-Is-Complete — Every upload tells the user what happened, once, in full, and shows the result where they dropped it

h3xxit:
Closes ticket: `Data/Engineering/Knowledge/Hexis/Tickets/Upload-Feedback-Is-Complete.md`

## Summary
- Fixed a silent-first-upload bug: `dispatchUpload` was awaiting `resolveSuggestionRouting` (ACL read, branch creation, workspace clone, change-request open) before touching any UI state, so the first suggestion-routed upload of a session left the tree blank for the whole await window (measured ~3.5s locally, longer on staging). The second upload short-circuited on an existing change request and answered instantly, which is why the bug looked intermittent.
- The upload notice and the error banner now render exactly once per page, scoped to the tree that received the drop.
- The error banner shows `Couldn't add <file name>`, the server's full reason wrapped across as many lines as needed (no truncation, no tooltip-only text), followed by a next-step line (retry, pick another folder, ask the owner).
- After a suggestion-routed upload, the proposed row appears in the tree without a reload, and the notice is now reliable on the first upload as well as the second.
- After an ordinary upload, the optimistic pending row is replaced by the real row once the tree refreshes, without a reload.

## Evidence
- Cause reproduced and timed on a local boot built from source: first suggestion-routed upload as a non-admin into `Skills/Engineering/coding/` — pre-UI await = 3555ms (listMyChangeRequests 8ms + createBranch 1669ms + getOrCreateWorkspace 1877ms) with nothing on screen; second upload = 112ms. See ticket ledger for full repro notes and the real server refusal text captured (`HTTP 403 ... Eligible: Admin.`).
- Fix pinned by a new test in `packages/core-frontend/src/modules/workspace/hooks/__tests__/useWorkspaceState.test.tsx` (`dispatchUpload: the drop is never silent`), verified RED against the pre-fix ordering before restoring the fix.
- Test suite: 467 files, 6928 passed, 1 skipped, 0 failed across all 4 packages (`NODE_ENV=test pnpm test`).
- Typecheck: clean across all 7 targets.
- Local boot from source (`deployment/docker-compose.build.yml`): `GET /api/health` → `{"status":"ok"}`.

---
## Summary by cubic
Uploads now report what happened, once, in the tree that received the drop: a progress notice appears from the first tick, refusals show the server's full reason, and the result lands in the tree without a reload.

**Bug Fixes**
- The first suggestion-routed upload of a session used to show nothing for seconds while a personal suggestions branch and its workspace were created; a progress notice naming the file and target folder now appears before that wait.
- A drop into `Skills/` used to paint the same banner above `Plugins/` as well, two in-flight drops could blank each other's notice, and a drop left a stale banner behind after a branch switch; banners are now stamped per tree, retired only when that tree has no upload running, and cleared when the branch changes.
- A refused upload used to truncate the server's reason behind a tooltip; the banner now wraps the full reason and ends with a next step, chosen from the refusal's HTTP status so reworded messages can't give the wrong advice.
- A suggestion-routed upload's proposed row now appears from the drop's own announcement and its notice says how to take it back; an ordinary upload's optimistic pending row is replaced by the real row once the refresh lands.

<sup>Written for commit 8573a4a55b756b107a3e8c46960ff0d501f5d17f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Bevel-Software/Hexis/pull/287?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


h3xxit:
All seven cubic findings on `c900e923` are addressed in code on `478ea5f6`. None were disputed.

**P2 — overlapping drops share one `uploadNotice`** (`workspace.context.ts`): valid, and worse than the comment says — the same slot problem applied to `uploadError`, where a drop into `Plugins/` wiped the error banner `Skills/` was still showing. Both are now keyed by the tree that took the drop (`uploadNotices` / `uploadErrors`, `ReadonlyMap<UploadTarget, …>`), and `clearUploadNotice` / `clearUploadError` take the tree they are clearing. Keying alone does not fix two drops into the *same* tree, so in-flight dispatches are counted per target and a tree's progress notice is retired only when that tree has none left running. Two new tests, both verified red against the old ordering: `keeps a second drop’s notice when the first settles` and `keeps the notice up while a sibling drop into the same tree is still running`, plus `shows both trees their own drop when two are in flight at once` in the sidebar test.

**P2 — progress notice survives a workspace switch** (`useWorkspaceState.ts`): valid. The stale upload rightly skips its own cleanup (`isCurrent()` is false, and its result belongs to the branch it was dispatched from), so the reset belongs with the workspace-change effect that already clears the optimistic overlays — both banner maps go with them. Pinned by `drops the banners of a branch the user has left`, which also asserts the stale upload settling afterwards does not put the notice back.

**P2 — `uploadErrorNextStep` classifies on backend prose** (`uploadError.ts`): valid. The signature is now `uploadErrorNextStep(status: number | undefined)`; `recordError` reads the status off the thrown `WorkspaceApiError` and `UploadError` carries it. 401/403 and 413 map onto the two specific steps, everything else (including a failure that never reached a response) gets the generic one. The hardcoded "50 MB" is gone — the cap is `MAX_UPLOAD_BYTES` in the backend and the server's own reason already carries it; a test asserts the advice contains no digits. Added `utils/__tests__/uploadError.test.ts`, and the banner test now uses refusal wording the old regexes would *not* have matched.

**P3 — `library:${dir}` spelled in four places** (`SkillsTree.tsx`): valid. `libraryUploadTarget(dir)` is exported beside `UploadTarget` and used by the tree and by both test files.

**P3 — vacuous "real row arrives" assertion** (`useWorkspaceState.test.tsx`): valid, and confirmed by experiment — with the post-upload `refreshFileTree()` stubbed out the old test still passed. `listFiles` is now a `mockImplementation` staggered on the commit: every read before the upload resolves answers without the file, the test asserts the path is absent at mount and again mid-flight (the optimistic `pendingUploads` entry is the only thing showing it), and the same stub-out now fails the test.

**P3 ×2 — `drop s` / `upload s own`**: fixed.

`pnpm typecheck` clean across all 7 targets; `NODE_ENV=test vitest run` in core-frontend: 183 files, 2472 passed, 0 failed. ESLint on the changed files is at the pre-change baseline (one pre-existing error in `FileExplorer.tsx:1569`, one pre-existing unused-directive warning).

h3xxit:
Second review pass. Head is now `8573a4a5`.

**Base sync.** `dev` had moved to `2654268e` (#288, withdraw a suggestion from the sidebar). Merged it in — two real conflicts, both in the suggestion-notice path #288 also touched:

- `useWorkspaceState.ts`: #288 added a sentence to the suggestion notice telling the user where the undo is; this branch had moved that notice into a per-tree `UploadNotice` object. Kept both — the object form now carries #288's fuller message verbatim, withdraw sentence included.
- `useWorkspaceState.test.tsx`: #288's assertion read `result.current.uploadNotice` as a string. Rewritten against `noticeIn(result.current)?.message`, keeping its `/To take it back, right-click the file and choose Withdraw suggestion\.$/` anchor so #288's copy stays pinned.

**The one finding open on `478ea5f6`** — the first `dispatchUpload` in `keeps the notice up while a sibling drop into the same tree is still running` running outside `act()`: valid and fixed in `8573a4a5`, as suggested. One correction to the finding's reasoning, for the record: no "not wrapped in act(...)" warning is actually emitted today — I ran that test and grepped for it before changing anything, and the count was zero. The change stands on its own terms regardless: the dispatch's prefix (`setUploadErrors`, `setUploadNotices`, `setIsUploading`) does run synchronously on the call, every other dispatch in the file wraps its start in `act`, and an un-acted synchronous state update is what starts warning the moment React's scheduling shifts. Re-verified that the test is still RED without the per-target in-flight counting it pins.

The other seven findings were addressed in `478ea5f6` — cubic has stamped five of them; the two it has not (`workspace.context.ts` overlapping drops, `SkillsTree.tsx` `library:` prefix) are both pinned to the old head with their anchors gone, and the fixes for both are described in the [previous comment](https://github.com/Bevel-Software/Hexis/pull/287#issuecomment-5745210233).

On `8573a4a5`: `pnpm typecheck` clean across all 7 targets; core-frontend `NODE_ENV=test vitest run` 183 files, 2479 passed, 0 failed (7 more than the last head — #288's tests arrived with the merge). ESLint on the changed file is clean.

Note for whoever reads the PR body: its Evidence section still quotes the first boot's numbers (`467 files`, `3555ms`) and predates both this merge and the review fixes.

PR #291, opened 2026-09-20
Tools-Refresh-After-Manual-Added — A tool manual added, changed or removed becomes visible and callable within a stated time, in the same connection (staging fixes)

h3xxit:
Ticket: `Data/Engineering/Knowledge/Hexis/Tickets/Tools-Refresh-After-Manual-Added.md`

Follow-up to #283 (already merged). Staging testing of the merged code failed three ways; this PR fixes them.

## Summary

- Applying a change request now emits `fs-tree-changed` for the default-branch workspace it rewrote. `mergePr` merges locally and pushes, so the following pull was a no-op and announced nothing, leaving tool and skill catalogs stale for up to 42s on staging.
- `hexis-mcp` gains a 2s heartbeat alongside the activity checks: an idle connection generates no activity, and `tools/list_changed` exists precisely for a client that is not asking.
- Catalog-check throttle drops from 5s to 2s and is counted from when a check started rather than settled (settle-stamping silently widened the window because a refresh runs inside a check).
- `AGENTS.md` states the five-second timeframe again and names the change-request commit path.

## Test plan

- `NODE_ENV=test pnpm test`: 493 files / 7439 passed / 1 skipped / 0 failures; typecheck and build clean.
- New: `merge-announces-tree.test.ts` (fails 2/5 with the fix reverted), idle-connection notification and heartbeat tests in `server.catalog-refresh.test.ts`, window-from-start tests in `catalog-watch.test.ts`.
- Live from-source compose run measured from the commit landing: deployment-side 154-398ms; idle local connection notified 375-2050ms; active connection 656-901ms; change-request merge 272-423ms.

---
## Summary by cubic
Fixes the staging failures that kept committed tool manuals and skills from appearing on existing MCP connections.

**Bug Fixes**
- A merge into the default branch now emits `fs-tree-changed` for the workspace it rewrote; the pull that follows is a no-op, so both applying a change request and an agent merging a branch had left approved manuals stale for tens of seconds.
- An idle `hexis-mcp` connection asks the deployment nothing — no heartbeat — and catches up at its next use; a connection in use sees a committed change within about five seconds.
- The catalog-check throttle drops from five seconds to two and is counted from when a check starts, so a slow refresh can't silently stretch the staleness window.

**Tests**
- Merge-announcement tests pin their own branch model, and a test pins that the announcement still fires after a failed post-merge pull, since the merge rewrote the workspace before the pull was attempted.

<sup>Written for commit 1b9b2431b9f427583c6931fca258d8f9d335f41c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Bevel-Software/Hexis/pull/291?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


h3xxit:
All five findings on `2a11fb31` are addressed (one commit, on top of a fresh merge of `origin/dev`).

**P2 — heartbeat above Node's timer ceiling.** Real, and nastier than it reads: the runtime does not clamp, it wraps, so `heartbeatMs: 30 days` becomes a check every millisecond — an embedding host asking for "hardly ever" would hammer its deployment forever. Clamped to `2_147_483_647` (clamped rather than refused: ~24.9 days is the longest this runtime can say, and that is what the caller meant). Test: a thirty-day heartbeat performs zero reads in 400ms, where a wrapped timer would have done hundreds.

**P3 — `CatalogCheck.check()`'s contract.** Correct, the doc was left behind by its own implementation. It now says the throttle counts from the last check's START, and why (a refresh runs inside a check, so counting from the finish adds the re-registration's seconds to every window).

**P3 — the merge tests and `DEFAULT_BRANCH`.** The stated mechanism is off by one detail — `packages/core-backend/vitest.config.ts` wires `setupFiles: ['./src/test-setup.ts']`, which calls `configureBranchModel(branchModelFromEnv())`, so `DEFAULT_BRANCH` is not `''` in the worker (if it were, `targetWorkspaceId` would be the empty string, the emit's own truthiness guard would skip, and test 1 would fail rather than pass). But the point stands and is the better one: the expectation was computed from the same binding the code reads, so it agreed with whatever the environment said and a hardcoded branch name could slip through wherever the two coincided. The file now pins its own model in `beforeAll` — to `trunk`, deliberately not `main`, so a hardcode fails here — restores the suite's model in `afterAll`, and asserts `DEFAULT_BRANCH === TRUNK` so the pin cannot silently stop applying.

**P3 — "within five seconds" versus the 15-second caveat.** Agreed, the strict bound made the neighbouring caveat a contradiction rather than a footnote. The sentence now carries the exception where the claim is made.

**P2 — announcing after a failed post-merge pull.** Kept, with the reasoning moved into the code. The tree the emit names is the one the merge ran in: `mergePr` resolves the target branch's own workspace and runs `git merge --no-ff` **there** before pushing, so the merged bytes are on that disk before the pull is even attempted — a pull that then fails cannot un-merge them, and what it failed to fetch is other people's commits, which leaves the catalogs no staler than they already were. (`baseBranch` is provably the branch the merge acted on: `preserveBaseRolesYaml` refuses the merge outright when it disagrees with the change request's own base.) A new test pins it: the pull throws, the announcement still fires for that workspace. If the merge were ever moved out of the target workspace into a temp clone, that test is where it would be caught.

`pnpm build` and `pnpm typecheck` clean; ESLint clean on every touched file; `pnpm test` with `NODE_ENV=test` on a fresh log: 500 files / 7508 tests / 1 skipped / 0 failures / 0 ENOSPC.

h3xxit:
Design change in 23fdd962, decided with Razvan: **the heartbeat is out.** An idle `hexis-mcp` connection asks the deployment nothing — a laptop nobody is using must not poll every two seconds for hours, and fifty of them must not do it together, which is the load #283's design refused on purpose. A connection learns of a change at its next use (a listing or a tool call finishing); on a connection in use that is within about five seconds of the commit. What stays from this PR: the merge announcement (the real staging bug — an applied change request never dropped the catalogs), the two-second throttle counted from a check's start, and the test that pins its own branch model. The guide, the README and the changeset now promise "at the connection's next use" rather than "within five seconds", and the ticket's idle-connection acceptance criterion should be read the same way. The timer-ceiling clamp and the heartbeat tests go with the timer.

---

PR #296, opened 2026-09-21
Tools-Refresh-After-Manual-Added — A tool manual added, changed or removed becomes visible and callable within a stated time, in the same connection (catalog-events stream)

h3xxit:
Ticket: `Data/Engineering/Knowledge/Hexis/Tickets/Tools-Refresh-After-Manual-Added.md`

Follow-up to #283 and #291 (both merged). Staging testing of the merged build left two failures: an idle `hexis-mcp` connection got no `tools/list_changed` (the 2s heartbeat was deliberately reverted for its per-laptop polling cost), and `list_tools` on an open connection landed at 5.0-6.4s.

## Summary

- New `GET /api/agent/catalog-events`: streams the caller's ACL-filtered catalog fingerprint (same digest as `/api/agent/catalog-revision`), one event on open and one each time it actually moves. Driven by `registerCatalogCacheInvalidation`'s new `onInvalidated` hook, so push and cache invalidation cannot disagree; silent when the fingerprint did not move, and bursts collapse onto one re-read. Advertised as a `catalogEvents` capability on `/api/config`.
- `hexis-mcp` subscribes once and feeds announcements to a new `CatalogCheck.notice()`, which shares the checker's applied-revision state, queues behind any in-flight refresh, and opens a throttle window so the provoked `tools/list` asks nothing.
- The heartbeat is not reinstated: an idle connection asks nothing on a timer, it parks one outbound socket. Activity checks remain the whole mechanism against a deployment that does not advertise the stream.
- `AGENTS.md` states the five-second timeframe again with the stream's caveat.

## Test plan

- `NODE_ENV=test pnpm test`: 505 files / 7645 passed / 1 skipped / 0 failures; build and typecheck clean.
- 33 new tests across core `catalog-events`, hexis-mcp `catalog-events`, `catalog-watch` `notice()`, `server.catalog-refresh` idle-connection integration, cache invalidation and public config.
- Live from-source compose run, timed from the commit landing: idle connection notified +180-217ms with zero catalog-revision polls (counted at a forwarder); active connection total 241-389ms; fallback with the stream blocked gives one stderr line and backoff reconnects, no hot loop.

---
## Summary by cubic
Adds a push-based catalog-change stream so long-lived `hexis-mcp` connections, idle or busy, get a new or changed tool manual or skill within five seconds without polling. This closes the gap where an idle connection heard nothing until its next use, and keeps active connections under the time budget.

**New Features**
- New `GET /api/agent/catalog-events` streams the caller's ACL-filtered catalog fingerprint, emitting once on open and again when it moves, driven by the same cache invalidation that drops catalog caches.
- `hexis-mcp` subscribes when the deployment advertises a `catalogEvents` capability, feeding announcements into the existing catalog checker with queueing and throttle windows.
- Activity checks remain the fallback for deployments that don't advertise the stream or proxies that won't carry it; the heartbeat is not reinstated.

<sup>Written for commit 091fbb674efb16ee31adaf76009766c6d6f9263d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Bevel-Software/Hexis/pull/296?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


h3xxit:
Closing without merging, by decision of Razvan Radulescu (2026-09-21).

The idle-connection half of acceptance criterion 2 is being amended rather than met with a new mechanism. The rule for client-side processes stands: a local `hexis-mcp` server checks the catalog when the connection is used and asks nothing while idle — no timer, and no parked socket per laptop either. An idle connection therefore catches up at its next use (the listing that runs the check, or the call after the one that triggered it); an active connection is within five seconds. The ticket, the guide and the changeset are being reworded to say exactly that, and the ticket goes back to staging testing against the amended criterion.

The deployment-side fix and the throttle correction from #291 are already on dev and stay. Nothing from this branch is needed for the amended criterion, so the stream is not landing; thank you for the measurements, they made the decision an informed one.

---

h3xxit:
All ten findings on `091fbb67` are addressed in code (one commit, on top of a fresh merge of `origin/dev` — which conflicted in `create-core-services.ts`, where dev's `pendingToolsService` and this branch's `catalogChanges` were added to the same returned object; both kept).

**P1 — CRLF frames.** Real, and it is the worst failure this file can have: the boundary scan looked for `\n\n`, a CRLF stream's boundary is `\r\n\r\n`, which contains no `\n\n` — so every frame accumulated until the 64KB cap tore the connection down, forever, without delivering one revision. A proxy that rewrites line endings is enough to produce it. The scan now matches every double line ending the spec allows (`\r\n\r\n`, `\r\n\n`, `\n\r\n`, `\n\n`, `\r\r`) and deliberately not `\r\n` alone, which is one ending rather than two; the line split inside a frame reads all three endings too. Test: a stub writing CRLF keep-alives and CRLF frames.

**P1 — the stream's 401.** Also real, and worse on an idle connection than it looks: this is the *only* request an idle connection makes, so nothing else was ever going to notice the bearer had expired and renew on its behalf — and a grant with no stated lifetime has no proactive timer either, so the stream reconnected forever with a retired key and the connection silently stopped hearing about changes. A 401 now goes through `renewConnectionKeyNow` (the single flight — never `config.renewConnectionKey` directly, since the refresh token rotates) and reopens once with the new bearer; a second 401 falls through to the backoff, where it is reported once. Key mode has no renewal and lands there directly. Test asserts the retry is immediate and carries the new bearer.

**P1 — an activity read overtaken by an announcement.** Correct, and the consequence you name is the bad one: applying the older fingerprint on top left `applied` pointing behind the catalog, so a refresh that failed halfway would leave the connection without tools and nothing recorded as owed. Announcements now bump an epoch; a read that sees it move discards what it read, because a fingerprint the deployment volunteered is by construction at least as fresh as one a read that started earlier is about to return. Test drives exactly that interleaving and asserts one refresh, no regression.

**P2 — a listing that waits only for `inFlight`.** Same race from the caller's side, same fix: `check()` now returns "everything learned so far, applied" — the check in flight, then the tail queued behind it. Two hops and no loop, so a workspace someone is committing to steadily cannot hold a listing open indefinitely. One existing test asserted promise *identity* between two concurrent `check()` calls, which this changes; it now asserts the property that mattered — one read, and neither caller answers before it settles.

**P3s.** `Both flags` → `These flags`. The stub's `announce()` skips a response that is no longer writable rather than trusting every caller to have emptied the set first. The README now states the capability gate (the stream is opened only against a workspace advertising `catalogEvents`; one that does not is not remarked on at all, and the checks are simply the whole mechanism there) and no longer offers "an older workspace" as a fallback example — you are right that such a deployment never gets a checker to fall back to. The reconnect test is renamed to what it verifies and now asserts the silence it describes (`expect(log).not.toHaveBeenCalled()`), so a regression that logged per reconnect fails it. The backend stream tests replace their 100–150ms sleeps with a waiter that first synchronises on the work a signal provokes — one catalog re-read, the event a debounce would push out — and then waits out a short quiet window; the hang-up test also gained a live-subscription control, so a version that simply never re-read would fail rather than pass twice.

`pnpm build` and `pnpm typecheck` clean; ESLint clean on every touched file; `pnpm test` with `NODE_ENV=test` on a fresh log: 514 files / 7814 tests / 1 skipped / 0 failures / 0 ENOSPC.

PR #300, opened 2026-09-21
Agents-File-Name-Configurable — The platform's agent guide can be given its own file name, so a customer's AGENTS.md is never overwritten

h3xxit:
Ticket: `Data/Engineering/Knowledge/Hexis/Tickets/Agents-File-Name-Configurable.md`

Adds a Deployment setting (knowledge-base section) for the agent guide's file name, default `AGENTS.md`. It appears in first-run setup and on the Deployment settings page, is restart-to-apply, and is validated (single `.md` name, not `CLAUDE.md`, distinct from platform files and root folder names).

- The managed guide is written and refreshed under the configured name on every protected branch; the template header names it.
- When renamed, a root `AGENTS.md` is ordinary content; a leftover platform-template `AGENTS.md` is removed on next start, anything else is untouched, with a boot note.
- `.bevelignore` rules follow the name; hand-written rules are kept.
- Agent-facing instructions name the configured file and also point at `AGENTS.md` if present.
- Platform-file rules (immovable/undeletable/read-through) and frontend tree confirmations follow the configured name.
- Optional pointer sentence appended to a customer-owned `AGENTS.md` (checkbox, on by default); never creates a file and never appends to a platform-written one.
- The three root folder settings and the new name lose their env-var overrides; `KB_KNOWLEDGE_BASE_DIR`, `KB_SKILLS_DIR`, `KB_PLUGINS_DIR` are imported once into saved settings with boot notes/warnings.
- Tests cover each case listed in the acceptance criteria.

---
## Summary by cubic
Lets deployments give the platform's managed agent guide its own file name, so a customer's `AGENTS.md` is never overwritten.

- Adds an **Agent guide file** deployment setting (default `AGENTS.md`) beside the root folders on first-run setup and the Deployment settings page; it validates (single `.md` name, not `CLAUDE.md`, distinct from platform files and folder names) and is restart-to-apply.
- With a custom name, the managed guide writes and refreshes under that name on every protected branch; a customer-owned `AGENTS.md` is left alone, while a leftover platform-written one is removed on next start with a boot note.
- `.bevelignore` rules, platform-file protections, and agent-facing instructions all follow the configured name; an optional checkbox (on by default) appends a pointer sentence to a customer's `AGENTS.md` without ever creating one.
- Retires `KB_KNOWLEDGE_BASE_DIR`, `KB_SKILLS_DIR`, and `KB_PLUGINS_DIR`: a value still set in the environment is imported once into the saved setting on first boot (with a log line naming the variable), and the saved value wins afterward.

<sup>Written for commit 2b5669b67057c085fc689429caa9e8107a87108a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Bevel-Software/Hexis/pull/300?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


h3xxit:
All eleven review findings are addressed in `32999c63`. What changed, finding by finding:

**P2 `seed-tree.ts` — two template entries, one destination.** A root template file named like the configured guide is now skipped when the template carries the guide source, so the managed entry wins deterministically rather than by walk order. Test: *lets the managed guide win a template that carries a root file of the same name*.

**P2 `deployment-settings.service.ts` — false restart notice.** `SettingDef` gained `unsetMeans`: what an unset row already means to the code that reads it. `save()` compares against that instead of a `DEFAULT_KB_LAYOUT` lookup, so first-saving `agentsFileLink: true` (already the default), or `agentsFile: AGENTS.md`, or a root's own default, owes no restart — while turning the box off still does. Test: *owes no restart for saving the value an unset setting already meant*.

**P2 `workspace.tools.ts` — stale descriptions after first-run setup.** Correct: setup applies the layout live, after the tools are mounted, so the catalog kept naming `AGENTS.md`. `configureKbLayout` now notifies listeners (`onKbLayoutApplied`), and `mount` rewrites each def's description in place when it fires. Descriptions that name the guide are functions rather than strings so they can be rebuilt. Rewritten in place rather than re-registered: the registry holds the object, both surfaces hold the same one, and provider-based registration would have moved every workspace tool behind the other static tools in the catalog. Test: *follows a layout applied after the tools were mounted*.

**P2 `kb-layout.ts` — the pointer sentence's link.** Agreed on the defect; solved differently from the suggestion, for a reason worth stating. The label is now backslash-escaped, but the destination is **not** percent-encoded — it is wrapped in angle brackets (`(<./Our [Agent] Guide (v2).md>)`) when a space or parenthesis would end it early. CommonMark §6.3 allows everything but `<`, `>` and a newline inside that form, none of which `validateFilename` admits, so the name survives verbatim. That matters: `reconcileLegacyGuide` decides whether to append by searching the customer's text for the guide's name, and a percent-encoded name would never be found there — every boot would append another copy, which is the one failure this feature exists to prevent. Test: *the pointer sentence links correctly for a name full of markdown punctuation*.

**P2 `template-files.step.ts:587` — positional provenance.** Right, `.gitattributes` above the rule is a coincidence anyone could hit. The bare rule now counts as the platform's only when the lines above it are the template's whole repo-hygiene block (its comment, `.gitignore`, `.gitattributes`), which still recognises every knowledge base seeded by a release that shipped the rule bare. Tests both ways: *keeps an operator's rule that merely happens to sit under .gitattributes* and *drops the bare rule an earlier release seeded in the template's hygiene block*.

**P2 `template-files.step.ts:395` — bypassing the injected filesystem.** Now `this.disk.readTextFile`, as suggested.

**P2 `template-files.step.ts:395` — symlink followed.** Now `this.disk.lstatOrNull` first, and only a regular file is classified at all: a link (or a directory under the name) is left untouched for both the removal and the pointer — writing through a link would have put our sentence in a file at the other end. A directory there also stops being an EISDIR boot failure. Test: *leaves a root AGENTS.md that is a symbolic link exactly where it points*.

**P3 tests — removal/keep asserted on one branch.** Both now loop `PROTECTED`, matching the sibling guide-refresh test.

**P3 changeset / P3 `.env.example` / P3 `UPGRADING.md`.** All three reworded. The changeset and `.env.example` now say plainly that only the three folder settings lose an override and that the guide's name never had a variable. `UPGRADING.md` no longer opens with a flat "nothing changes": it names the one deployment shape that does change (a saved value differing from a still-set `KB_*`), says the saved value now wins where the variable used to, and says that editing `.env` cannot hand precedence back because the import only runs where nothing is saved.

Verification: `pnpm typecheck` clean; `packages/core-backend` 292 files / 4773 tests pass. The `core-frontend` suite fails in this workspace with `React.act is not a function` on an unmodified checkout of `2b5669b6` too — a local environment fault, not this branch. Lint output is byte-identical to the baseline (170 pre-existing problems).

h3xxit:
The one finding cubic still has open — `kb-layout.ts:199`, the pointer sentence — is now fixed properly in `a87f16b1`. My previous answer argued for angle brackets over percent-encoding; that was half right, and the half it got wrong is the half that matters.

Angle brackets do handle a space and a parenthesis. They do **not** handle the two characters that actually break the target, and both are legal in a guide name (`validateFilename` forbids `<>:"/\|?*` and control characters, not these):

- `#` opens a URL fragment. `[x](<./#2 Guide.md>)` resolves to `./` with a fragment — a link to the customer's own file, which is exactly the failure the finding named.
- `%` is a malformed percent-escape unless encoded. `./100%.md` is not a valid path.

So the destination is percent-encoded now: `encodeURI`, plus `#`, `(` and `)` escaped by hand because `encodeURI` leaves all three and each ends the path early. The label keeps its backslash escaping.

The objection I raised against encoding was real, so it is fixed rather than traded away. `reconcileLegacyGuide` decided whether to append by searching the customer's text for the raw name; once the sentence spells the name escaped in the label and encoded in the destination, a punctuated name never appears raw in the file we wrote, and every boot would append another copy. That question is now `mentionsAgentsFile(text, name)` in the layout module, which accepts the plain name **or** either spelling the platform itself writes. The customer's own mention in their own words still counts, so the acceptance criterion reads the same.

New tests: the encoded destination for `Our [Agent] Guide (v2).md`, for `#2 Guide.md` and for `100%.md`; that a file already carrying the sentence reads as mentioning the guide (and that a silent file does not); and a two-boot `TemplateFilesStep` test asserting a punctuated name is appended once, not once per start — that one fails against the previous commit.

Verification on `a87f16b1`: `pnpm typecheck` clean, `packages/core-backend` 292 files / 4776 tests pass, lint byte-identical to the baseline (170 pre-existing problems). The `core-frontend` suite still fails in this workspace with `React.act is not a function`, reproduced on an unmodified checkout — a local environment fault, not this branch.

h3xxit:
Both findings from the review of `a87f16b1` are fixed in `94ea28d2`.

**P2 `template-files.step.ts:434` — a second rename appended instead of updating.** Correct, and it is the more interesting half of the same bug I fixed last round. Renaming the guide `HEXIS.md` → `GUIDE.md` left the customer's `AGENTS.md` carrying a sentence pointing at a file that no longer exists, and put a second one underneath it: the guard asks whether the NEW name is mentioned, and the old sentence does not mention it.

The sentence is now recognisable by its shape rather than by the name it happens to carry. `agentsFilePointerSentence` was split so the name-independent tail is a constant, and `retargetAgentsFilePointer(text, agentsFile)` matches on that fixed text at both ends with the label and destination in between — the label allowing the backslash escapes `markdownLinkLabel` writes, the destination unable to hold a `)` because `agentsFileLinkPath` encodes it. Found and different, it is re-aimed and the note says `Point the sentence in AGENTS.md at GUIDE.md`; found and already right, nothing is written; not found at all, the append path runs as before against the customer's own text. It returns the text unchanged rather than null in the "ours and correct" case precisely so the caller can tell that apart from "none of ours here".

One deliberate ordering note: the consent checkbox still gates every write, so a deployment with the box unticked gets nothing written — stale sentence or not. That is what "unticking the box is how an admin says no" has to mean, and the criterion says the platform writes only while the box is on.

**P3 `steps.test.ts:979` — comment and assertion disagreed.** Right on both counts: the phrase counter proved no second sentence was appended, which the byte comparison above it had already settled, and it said nothing about committing. Replaced with the `git dir` pattern the sibling tests use — HEAD captured before the second boot and asserted unchanged after — so the comment now states what is checked.

New coverage: `retargetAgentsFilePointer` from three previous names (including `Our [Agent] Guide (v2).md` and `#2 Guide.md`), a text already aimed correctly, and three texts holding nothing of ours (plain prose, the customer's own link, and our opening words inside their own sentence). Plus a startup test that renames twice and asserts exactly one sentence, naming the guide that exists, with the note.

Verification on `94ea28d2`: `pnpm typecheck` clean; `packages/core-backend` 292 files / 4779 tests pass; lint byte-identical to the baseline (170 pre-existing problems). `core-frontend` still fails in this workspace with `React.act is not a function`, reproduced on an unmodified checkout — a local environment fault, not this branch.

